PR #1, opened 2019-05-25
Refactor policy file and service generation

tarokkk:
- Generate hosts instead of wildcard host definition
- Add checksum to proxy policy change (automatic redeploy of pods on configuration change)


desimone:
Thank you!

PR #5, opened 2019-05-30
v0.0.5 fixes

### LINKED ISSUE #3: Couldn't find key idp-service-account
idp-service-account is an optional identity provider configuration variable but currently our helm chart will fail if no value is provided.

### LINKED ISSUE #4: improper indentation for resources
Replication steps:
```bash
helm install $HOME/pomerium-helm \
	--set service.type="NodePort" \
	--set ingress.secret.name="pomerium-tls" \
	--set ingress.secret.cert=$(base64 -i "$HOME/.acme.sh/*.corp.beyondperimeter.com_ecc/fullchain.cer") \
	--set ingress.secret.key=$(base64 -i "$HOME/.acme.sh/*.corp.beyondperimeter.com_ecc/*.corp.beyondperimeter.com.key") \
	--set config.policy="$(base64 -i docs/docs/examples/config/policy.example.yaml)" \
	--set authenticate.idp.provider="google" \
	--set authenticate.idp.clientID="XXXX" \
	--set authenticate.idp.clientSecret="XXXXX" \
	--set-string ingress.annotations."kubernetes\.io/ingress\.allow-http"=false \
	--set ingress.annotations."cloud\.google\.com/app-protocols"=\"{\"https\":\"HTTPS\"}\"
```
Error:

```
Error: YAML parse error on pomerium/templates/authenticate-deployment.yaml: error converting YAML to JSON: yaml: line 105: did not find expected key
```

Expected:

Working helm deployment

Environment: Google GKE

desimone:
Fixes #3 
Fixes #4 (thx @travisgroth ! https://github.com/desimone/pomerium-helm/pull/1) 

- [x] confirm everything works on GKE
- [x] ready for review

### DIFF
--- a/templates/authenticate-deployment.yaml
+++ b/templates/authenticate-deployment.yaml
@@ -57,6 +57,8 @@ spec:
         env:
         - name: SERVICES
           value: authenticate
+        - name: AUTHENTICATE_SERVICE_URL
+          value: {{ default (printf "https://authenticate.%s" .Values.config.rootDomain ) .Values.proxy.authenticateServiceUrl }}
         - name: COOKIE_SECRET
           valueFrom:
             secretKeyRef:
@@ -67,10 +69,6 @@ spec:
             secretKeyRef:
               name: {{ $secretName }}
               key: shared-secret
-        - name: REDIRECT_URL
-          value: {{ default (printf "https://%s.%s/oauth2/callback" (include "pomerium.authenticate.fullname" .)  .Values.config.rootDomain ) .Values.authenticate.redirectUrl }}
-        - name: PROXY_ROOT_DOMAIN
-          value: {{ .Values.config.rootDomain }}
         - name: IDP_PROVIDER
           value: {{ .Values.authenticate.idp.provider }}
         - name: IDP_CLIENT_ID
@@ -85,11 +83,13 @@ spec:
               key: idp-client-secret
         - name: IDP_PROVIDER_URL
           value: {{ .Values.authenticate.idp.url }}
+{{- if .Values.authenticate.idp.serviceAccount }}
         - name: IDP_SERVICE_ACCOUNT
           valueFrom:
             secretKeyRef:
               name:  {{ $secretName }}
               key: idp-service-account
+{{- end }}
         - name: CERTIFICATE
           valueFrom:
             secretKeyRef:
@@ -132,8 +132,6 @@ spec:
         configMap:
           name: {{ $configName }}
 {{- end }}
-      resources:
-{{ toYaml .Values.resources | indent 10 }}
 {{- if .Values.extraVolumes }}
       volumes:
 {{- toYaml .Values.extraVolumes | indent 8 }}
--- a/templates/proxy-deployment.yaml
+++ b/templates/proxy-deployment.yaml
@@ -71,11 +71,9 @@ spec:
         - name: AUTHENTICATE_SERVICE_URL
           value: {{ default (printf "https://authenticate.%s" .Values.config.rootDomain ) .Values.proxy.authenticateServiceUrl }}
         - name: AUTHORIZE_SERVICE_URL
-          value: {{ default (printf "https://authorize.%s" .Values.config.rootDomain ) .Values.proxy.authorizeServiceUrl }}
+          value: {{ default (printf "https://%s.%s.svc.cluster.local" (include "pomerium.authorize.fullname" .) .Release.Namespace ) .Values.proxy.authorizeInternalUrl}}
         - name: AUTHENTICATE_INTERNAL_URL
-          value: {{ default (printf "%s.%s.svc.cluster.local" (include "pomerium.authenticate.fullname" .) .Release.Namespace ) .Values.proxy.authenticateInternalUrl}}
-        - name: AUTHORIZE_INTERNAL_URL
-          value: {{ default (printf "%s.%s.svc.cluster.local" (include "pomerium.authorize.fullname" .) .Release.Namespace ) .Values.proxy.authorizeInternalUrl}}
+          value: {{ default (printf "https://%s.%s.svc.cluster.local" (include "pomerium.authenticate.fullname" .) .Release.Namespace ) .Values.proxy.authenticateInternalUrl}}
         - name: CERTIFICATE
           valueFrom:
             secretKeyRef:
@@ -118,8 +116,6 @@ spec:
         configMap:
           name: {{ $configName }}
 {{- end }}
-      resources:
-{{ toYaml .Values.resources | indent 10 }}
 {{- if .Values.extraVolumes }}
       volumes:
 {{- toYaml .Values.extraVolumes | indent 8 }}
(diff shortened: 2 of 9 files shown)


desimone:
/cc @travisgroth two issues I'm stuck on and welcome any ideas. 

1.  Any ideas how to pass the GKE service annotation `cloud.google.com/app-protocols: '{"https":"HTTPS"}' `  with the proper escapes so can pass those details in via the command line instead of with values?
2. I was hitting an error with the resources key. You can see it is removed above. I couldn't get it working.

travisgroth:
1.  I think this will do it.  

```shell
helm template . \
--set ingress.enabled=true \
--set ingress.annotations."cloud\.google\.com/app-protocols"='\{"https":"HTTPS"\}
```
```yaml
# Source: pomerium/templates/ingress.yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: release-name-pomerium
  labels:
    app.kubernetes.io/name: pomerium
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Tiller
    helm.sh/chart: pomerium-1.0.0
  annotations:
    cloud.google.com/app-protocols: '{"https":"HTTPS"}'
    kubernetes.io/ingress.allow-http: "false"
```
Funny thing here.  The quotes are only needed for one of the 4 ingresses:

```shell
helm template . --set ingress.enabled=true --set ingress.annotations.cloud\.google\.com/app-protocols='\{"https":"HTTPS"\}'  | grep cloud
    cloud.google.com/app-protocols: '{"https":"HTTPS"}'
    cloud.google.com/app-protocols: '{"https":"HTTPS"}'
    cloud.google.com/app-protocols: '{"https":"HTTPS"}'
    cloud:
```

travisgroth:
2. The resource block needs to move over two spaces.  https://github.com/desimone/pomerium-helm/pull/1

3.  It looks like there will also be a conflict between the config volume and extravolumes if you specify both right now.  I don't have time to reason through the change right now, but it is unlikely to be hit in the near term.

templates/proxy-deployment.yaml
```diff
@@ -71,11 +71,9 @@ spec:
         - name: AUTHENTICATE_SERVICE_URL
           value: {{ default (printf "https://authenticate.%s" .Values.config.rootDomain ) .Values.proxy.authenticateServiceUrl }}
         - name: AUTHORIZE_SERVICE_URL
-          value: {{ default (printf "https://authorize.%s" .Values.config.rootDomain ) .Values.proxy.authorizeServiceUrl }}
+          value: {{ default (printf "https://%s.%s.svc.cluster.local" (include "pomerium.authorize.fullname" .) .Release.Namespace ) .Values.proxy.authorizeInternalUrl}}
         - name: AUTHENTICATE_INTERNAL_URL
```
victornoel: @desimone this should be removed (if it wasn't put back since v0.0.4)

templates/proxy-deployment.yaml
```diff
@@ -71,11 +71,9 @@ spec:
...
           value: {{ default (printf "%s.%s.svc.cluster.local" (include "pomerium.authenticate.fullname" .) .Release.Namespace ) .Values.proxy.authenticateInternalUrl}}
-        - name: AUTHORIZE_INTERNAL_URL
```
victornoel: @desimone I think REDIRECT_URL and ROOT_DOMAIN should be removed as well since v0.0.4

templates/proxy-deployment.yaml
```diff
@@ -71,11 +71,9 @@ spec:
         - name: AUTHENTICATE_SERVICE_URL
           value: {{ default (printf "https://authenticate.%s" .Values.config.rootDomain ) .Values.proxy.authenticateServiceUrl }}
         - name: AUTHORIZE_SERVICE_URL
-          value: {{ default (printf "https://authorize.%s" .Values.config.rootDomain ) .Values.proxy.authorizeServiceUrl }}
+          value: {{ default (printf "https://%s.%s.svc.cluster.local" (include "pomerium.authorize.fullname" .) .Release.Namespace ) .Values.proxy.authorizeInternalUrl}}
```
victornoel: See https://github.com/pomerium/pomerium/issues/153 by the way

desimone: Should be fixed by https://github.com/pomerium/pomerium/pull/154 thx!

desimone:
\cc @victornoel @travisgroth if you don't mind giving this another look. This is what I'll merged (see updated tag) after v0.0.5 drops.

victornoel:
@desimone if you can wait I will test this version of the chart with the latest build on Monday

desimone:
Absolutely

victornoel:
@desimone there is something strange with `.Values.config.existingPolicy`: it seems it is referenced in some place but it is never really used. I think maybe it should be completely removed?

templates/proxy-deployment.yaml
```diff
@@ -71,11 +71,9 @@ spec:
...
         - name: AUTHENTICATE_INTERNAL_URL
           value: {{ default (printf "%s.%s.svc.cluster.local" (include "pomerium.authenticate.fullname" .) .Release.Namespace ) .Values.proxy.authenticateInternalUrl}}
```
victornoel: @desimone shouldn't this one be prefixed with https:// since https://github.com/pomerium/pomerium/pull/154?

travisgroth:
>@desimone there is something strange with .Values.config.existingPolicy: it seems it is referenced in some place but it is never really used. I think maybe it should be completely removed?

I must've introduced that while reworking the chart.  I agree - I don't believe it makes sense anymore in the current state of things.  An existing CM for just policy wouldn't make much sense.  (1) config.existingConfig covers that and (2) config.policy is inlined to the in-chart CM.

templates/authenticate-deployment.yaml
```diff
@@ -85,11 +83,13 @@ spec:
               key: idp-client-secret
         - name: IDP_PROVIDER_URL
           value: {{ .Values.authenticate.idp.url }}
+{{- if .Values.authenticate.idp.serviceAccount }}
```
victornoel: @desimone this prevents the use of a service account when rolling my own secret containing the service account.

As a workaround, I can just set `authenticate.idp.serviceAccount` to `true` though. I'm not sure how to fix this so we may need to leave it like this but simply document the fact that if needed, it can be set to `true` when using `existingSecret`.

templates/proxy-deployment.yaml
desimone: Yes, good catch.

victornoel:
@desimone I think when you force-pushed yesterday you undid some of the change you had done: https://github.com/pomerium/pomerium-helm/compare/cb1b10c89fa9c49cfa2d0de7a6d608a587cf0ee2..95d2757b184bc0a8af3c0b597beaed890df78b3f

desimone:
@victornoel Thank you. I totally did.

templates/authenticate-deployment.yaml
desimone: Lets track this in a new issue, #7

PR #9, opened 2019-06-06
Fix config checksum

victornoel:
This definitively fix the way the checksum is done to force reload the pods on change:
- it was incorrectly applied to the deployment annotations
- it had an incorrect name
- it wasn't based on files

I also added the same for the secrets to be consistent.

I just wonder how all of this relates to hot-reloading:
- for sure for the secrets it is needed since they ends up in the pod as environment variables, so hot-reloading is not related
- but the config itself, if there is hot reloading maybe it is not needed to reload the pod?

### DIFF
--- a/templates/authenticate-deployment.yaml
+++ b/templates/authenticate-deployment.yaml
@@ -23,7 +23,8 @@ spec:
   template:
     metadata:
       annotations:
-        config/checksum: {{ print .Values.config.extraOpts | sha256sum }}
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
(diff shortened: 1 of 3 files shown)


travisgroth:
Hey @victornoel thanks for this.  

>for sure for the secrets it is needed since they ends up in the pod as environment variables, so hot-reloading is not related

Yes; correct.  That was omitted in the first version for reasons that escape me now.  

>but the config itself, if there is hot reloading maybe it is not needed to reload the pod?

Only policy is reloaded on the fly presently.  If you configure other options, it would need the checksum based reload.  IIRC, this is the reason that only extraOpts was checksummed.  I think that's still the right thing to do until a full process reinitialize with all the options is supported.

However, adding the secrets checksum and (obviously) relocating the misplaced checksums makes sense to me.

PR #11, opened 2019-06-13
Fix config checksum

victornoel:
Rework of #9 using only extraOpts for the config checksum and whole file for the secrets.

/cc @travisgroth @desimone


desimone:
Thanks!

PR #14, opened 2019-07-25
Resources block must be at container level not pod

aaron-trout:
The helm chart as of commit b09e4b7fce2be4838c1f5e5c16553bc3800b8a72 does not work because the resources block was added at the pod level instead of the container level where it belongs. Just moved these lines up a bit and indented `resources:` one level to make it part of the container spec rather than pod spec (the `indent 10` was actually already correct).

Such YAML, many Devops, wow!

### DIFF
--- a/templates/authenticate-deployment.yaml
+++ b/templates/authenticate-deployment.yaml
@@ -128,6 +128,8 @@ spec:
             path: /ping
             port: https
             scheme: HTTPS
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
 {{- if or .Values.config.existingConfig .Values.config.policy }}
         volumeMounts:
         - mountPath: /etc/pomerium/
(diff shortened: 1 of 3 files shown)


desimone:
Thanks @aaron-trout !

Looks good to me but I'm far from a helm expert. 

\cc @victornoel  @travisgroth 

> Such YAML, many Devops, wow!

Modern problems require modern solutions. 😉 

0lg04ovga0m11

victornoel:
@desimone ha, sorry about that, it was my fault :)

The chart on helm stable repository has the same problem btw

travisgroth:
@aaron-trout LGTM.  Would you mind updating your commit to include [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md) line so I can directly apply this patch into helm/stable?  

Thanks!

aaron-trout:
Cool, have done that @travisgroth! Didn't realise it was in the official charts repo actually, I was just using it direct from this git repo :)

travisgroth:
Yep we're keeping them in sync.  Thanks for the contribution!

PR #17, opened 2019-07-31
Add ServiceMonitor template for Prometheus Operator

travisgroth:


desimone:
@travisgroth should we do something similar for Jaeger/tracing?

travisgroth:
>should we do something similar for Jaeger/tracing?

Yes, I think so.  The more options we expose directly, the easier the chart is to use without guesswork.

PR #32, opened 2019-11-01
Setting the Hosts explicitly

mohsen0:
The `kubernetes-sigs/external-dns` create DNS records based on the Hostnames in Ingress objects.
Setting the hostnames explicitly in the Ingress object and not using a wildcard can help to take advantage of this functionality


mohsen0:
H @desimone 
Can this PR be reviewed? 
Thanks
M

README.md
```diff
@@ -80,32 +80,32 @@ You may force recreation of your TLS certificates by setting `config.forceGenera
...
 1) turn `generateTLS` to `false`
-2) specify `authenticate.existingTLSSecret`, `authorize.existingTLSSecret`, and `proxy.existingTLSSecret`, pointing at the appropriate TLS certificate for each service.  
+2) specify `authenticate.existingTLSSecret`, `authorize.existingTLSSecret`, and `proxy.existingTLSSecret`, pointing at the appropriate TLS certificate for each service.
 
 All services can share the secret if appropriate.
 
 ## Configuration
 
-A full listing of Pomerium's configuration variables can be found on the [config reference page](https://www.pomerium.io/docs/config-reference.html).
+A full listing of Pomerium's configuration variables can be found on the [config reference page](https://www.pomerium.io/docs/reference/reference.html).
```
desimone: Thanks, good catch.

templates/ingress.yaml
```diff
@@ -16,26 +16,30 @@ spec:
   tls:
     - secretName: {{ default .Values.ingress.secretName .Values.ingress.secret.name}}
       hosts:
-        - {{ printf "*.%s" .Values.config.rootDomain | quote }}
```
desimone: I have mixed feelings about ditching the wildcard domain here. This would result in each managed host needing a change not only to policy, but also to the ingress. Thoughts @travisgroth ?

travisgroth:
Hey @mohsen0 sorry for not seeing this sooner.  Thanks.

Hmm we just approved a PR that took the opposite approach in helm/stable.  However, when we add route scoped sessions https://github.com/pomerium/pomerium/pull/373, we will have to refactor this regardless.  

Looking forward, the ingress hosts list could be generated from the policy structure's `from` fields, and then tack on the authenticate service name.  I think.  This would make the list explicit but not repeated in two data structures.  

@mohsen0 thoughts on that approach instead?

templates/ingress.yaml
mohsen0: I understand some users might choose to manage the DNS records in another way. But proxy already mandates to have the list of `config.policy.[].from`.

mohsen0:
To be honest, Before raising this PR, I did try to implement the same approach, but I gave up :) .
So I fall back on a simple approach, I will spend time again on this later today and I will let you know.


values.yaml
```yaml
config:
  rootDomain: example.com
  policy:
  - from: https://prom.example.com
    to: https://prometheus.prod-eu-west-1.example.com
    timeout: 30s
  - from: https://prom.au.example.com
    to: https://prometheus.prod-ap-southeast-2.example.com
    timeout: 120s
```

In templates/helper.tpl
```
{{- define "mychart.hosts" -}}
{{- /*  */ -}}
{{- $hosts := list -}}
{{- range $.Values.config.policy }}
{{  $hosts := append $hosts .from }}
{{- end }}
{{- $hosts }}
{{- end -}}
```
In templates/ingress.yaml
```
{{- $hosts := include "mychart.hosts" $ -}}
{{- if .Values.ingress.enabled -}}
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: my-chart-name
spec:
  rules:
    {{- range $hosts }}
    - host: {{ printf "%s.%s" . $.Values.config.rootDomain }}
      http:
        paths:
          - paths:
            backend:
              serviceName: my-serviceName
              servicePort: http
    {{- end }}
{{- end }}
```

```
 templates/ingress.yaml:22:16: executing "templates/ingress.yaml" at <$hosts>: range can't iterate over
```

mohsen0:
Hi @travisgroth / @desimone 
That is ready to be reviewed

mohsen0:
I will update the stable/pomerium when this gets merged.

travisgroth:
I was tinkering with this last night and I think we both arrived at the same approach.  Just for posterity, this was what I wound up with (though not complete):

```yaml
spec:
  tls:
    - secretName: {{ default .Values.ingress.secretName .Values.ingress.secret.name}}
      hosts:
        {{- range .Values.config.policy }}
        - {{ trimPrefix "https://" .from }}
        {{- end }}
        {{- if not .Values.service.authorize.headless }}
        - {{ printf "authorize.%s" .Values.config.rootDomain }}
        {{- end }}
        - {{ printf "authenticate.%s" .Values.config.rootDomain }}
  rules:
    {{- range .Values.config.policy }}
    - host: {{ trimPrefix "https://" .from }}
      http:
        paths:
          - paths:
            backend:
              serviceName: {{ template "pomerium.proxy.fullname" $ }}
              servicePort: https
    {{- end }}
```

This is a behavior change from wildcard on a domain.  Things to consider:

1. A way to opt in or out of the new behavior.  The authorize/authenticate endpoint still need a rootDomain equivalent.  

If we set this to default on, I think this is at least a minor version change (4.1.0) as it does alter the internals a bit.  I'm worried what a removal of the default wildcard would do to the TLS configuration in particular.  People may be leaning on that and I'm not sure what happens when the host list is inconsistent with the certificate Subject list.  It may vary by ingress controller.  Perhaps we mark this 5.0.0 to be safe and just default on? 

2. An option to set an explicit host list instead.  This could be an add-on parameter that augments the current rootDomain parameter.  Non-exhaustive reasons:
  a. you have many policies and don't want N ingress routes (scale)
  b. you have certs coming in elsewhere and/or want wildcard host entries and wish to take advantage of the new multi-domain support

There might be a way to reconcile those in a single parameter.  Perhaps something like `ingress.hosts` could default to an empty list and enumerate the policy list if empty.  If it contains hosts, use those to override the list generated from the policy list.  Thoughts?  

re: syncing with helm/stable, I'll can take care of it.  I will probably take this patch there first to have their CI validate.  If you want to pivot to a PR there first, that's also okay.  Sorry for the mess between the two repos.  I think we're going to move to Hub soon but need to get our own CI setup going first.  

Thanks again for the contribution.

mohsen0:
So About the Ingress certificate, I generated a wild card certificate for `*.{{.Values.config.rootDomain }}`.
and I believe proxy and authenticator do the same thing.

travisgroth:
>So About the Ingress certificate, I generated a wild card certificate for *.{{.Values.config.rootDomain }}.
and I believe proxy and authenticator do the same thing.

I'm seeing the same thing.  And that does match the API spec.  

Would you mind adding in the ability to explicitly set the ingress hosts and then bump the chart version to 4.1.0?

mohsen0:
When this one gets merged in, I like to raise the other PR :)

mohsen0:
Can you approve this too?

desimone:
Thanks for the contribution @mohsen0

Issue #34, opened 2019-12-06
Helm lint --strict fails


Dragotic:
Hello guys,

So, I was playing with Pomerium using the helm charts. I change the configuration to fit my deployment needs, but when I tried to deploy the chart it failed.

I have a `helm lint --strict` stage in my pipeline which failed the deployment.

```
➜  pomerium helm lint --strict pomerium                                                                  
==> Linting pomerium
[ERROR] templates/: render error in "pomerium/templates/tls-secrets.yaml": template: pomerium/templates/_helpers.tpl:5:31: executing "pomerium.name" at <.Values.nameOverride>: map has no entry for key "nameOverride"

Error: 1 chart(s) linted, 1 chart(s) failed

➜  pomerium helm lint --strict pomerium                                                                  
==> Linting pomerium
[ERROR] templates/: render error in "pomerium/templates/secret.yaml": template: pomerium/templates/secret.yaml:1:18: executing "pomerium/templates/secret.yaml" at <.Values.config.exist...>: map has no entry for key "existingSecret"

Error: 1 chart(s) linted, 1 chart(s) failed
```

Some of these configurations are not present in the default values.yaml file, nor are they present in the [Readme](https://github.com/pomerium/pomerium-helm#configuration)

Dragotic:
This was fixed in #36 PR.

/Close

PR #36, opened 2019-12-09
Add missing values to values.yaml

Dragotic:
This is a suggestion that came up, upon deploying this chart on a sandbox cluster of mine.

As seen in #34 this PR will fix the errors from `helm lint --strict`


desimone:
Looks reasonable to me, thank you for the PR @Dragotic .

values.yaml
```diff
@@ -1,13 +1,17 @@
...
+nameOverride: pomerium
+fullnameOverride: pomerium
+
 # settings that are shared by all services
 config:
   # routes under this wildcard domain are handled by pomerium
   rootDomain: corp.beyondperimeter.com
-  # existingSecret:
-  # existingConfig:
+  existingSecret: ""
```
desimone: AFAIK, empty strings are equivalent to [being nil in go templates](https://golang.org/pkg/text/template/#hdr-Actions).

values.yaml
```diff
@@ -1,13 +1,17 @@
 # For detailed explanation of each of the configuration settings see
 # https://www.pomerium.io/reference/
 
+nameOverride: pomerium
```
desimone: I'm not sure we want to override the name / full name in it's default configuration. Is this required for passing strict?

values.yaml
```diff
@@ -17,8 +21,8 @@ config:
   policy: {}
 
 authenticate:
-  # fullnameOverride: authenticate
-  # nameOverride: authenticate
+  fullnameOverride: authenticate
```
desimone: Same comment as for `name/fullname`

values.yaml
```diff
@@ -125,6 +129,9 @@ extraEnv: {}
 extraArgs: {}
 extraVolumes: {}
 
+annotations: {}
+imagePullSecrets: ""
```
desimone: Can you provide a little more context on `imagePullSecrets`? Is this required for lint'ing or?

values.yaml
Dragotic: Yes, that's correct. Any empty string will result in an empty pipeline value which doesn't generate output.

values.yaml
Dragotic: Yes, unfortunately, it is required when running `helm lint --strict`.

We can change it to `""` if you think it's better.

values.yaml
Dragotic: Yes, unfortunately, it is required when running `helm lint --strict`.

We can change it to `""` if you think it's better.

values.yaml
Dragotic: Same here it is required for lint'ing.

Dragotic:
Yes, I will update the README.

desimone:
> **Update:** I have updated the README. Please note that I didn't add the following to the README, because they are documented on the `values.yaml` and are the default for every helm chart with no difference in the Pomerium Chart:
> 
> ```yaml
> affinity: {}
> tolerations: []
> nodeSelector: {}
> ```

Perfect. I think that makes sense. 

> Also, check that the following descriptions match:
> 
> ```lisp
> config.existingSecret: Name of the existing Kubernetes Secret.
> config.existingConfig: Name of the existing Config Map deployed on Kubernetes.
> config.existingCASecret: Name of the existing CA Secret.
> ```

👍 

LGTM. Just a heads up, might be a few days before this gets merged until I get back home

travisgroth:
@Dragotic Thanks for the PR.  Could you either (a) DCO sign these commits so I can sync this to helm/stable, or (b) if you'd like, mirror your PR there with DCO sign off?

Dragotic:
@travisgroth No problem! I signed-off the two commits. Tell me if something else is missing.

values.yaml
travisgroth: See other comment on overrides.

values.yaml
```diff
@@ -17,8 +21,8 @@ config:
   policy: {}
 
 authenticate:
-  # fullnameOverride: authenticate
-  # nameOverride: authenticate
+  fullnameOverride: authenticate
+  nameOverride: authenticate
```
travisgroth: See other comment on overrides.

values.yaml
```diff
@@ -33,31 +37,31 @@ authenticate:
...
     defaultIPList: []
-  # replicaCount: 1
+  replicaCount: 1
 
 
 authorize:
-  # fullnameOverride: authorize
-  # nameOverride: authorize
+  fullnameOverride: authorize
+  nameOverride: authorize
```
travisgroth: See other comment on overrides.

values.yaml
```diff
@@ -33,31 +37,31 @@ authenticate:
...
     defaultIPList: []
-  # replicaCount: 1
+  replicaCount: 1
 
 
 proxy:
-  # fullnameOverride: proxy
-  # nameOverride: proxy
+  fullnameOverride: proxy
+  nameOverride: proxy
```
travisgroth: See other comment on overrides.

values.yaml
victornoel: yes I think so too, I made the same comment on the helm repository actually :)

Dragotic:
@victornoel @travisgroth Thanks for the comments guys. Check again for the new commits.

travisgroth:
Merged in https://github.com/helm/charts/pull/19527.  👍

Issue #35, opened 2019-12-09
service annotations


derrickburns:
Please modify your helm chart to allow distinct annotations for each service and deployment.

There are several use cases where being able to provide different annotations is useful, mainly having to do with secrets:

- reloader
- gloo gateway

desimone:
Hi @derrickburns ,

Could you elaborate / link to some similar configurations so I can get a sense of what you had in mind?

Thanks!

derrickburns:
Ah.  Ok.

Reloader is a service that restart pods in a deployment when a secret or configmap changes.  You configure reloader by adding annotations to a deployment.  I would like to automate the restart of pomerium deployments if the secrets or configmaps they depend on change.  Right now, you have different secrets for each deployment.  Hence, we need different annotations.

Similarly, the gloo gateway allows one to configure "upstreams" as requiring TLS.  This can be done my a gloo specific CRD or via an annotation on a service.  Again, if we use different TLS secrets for the different pomerium services, we need different annotations.

desimone:
@derrickburns I think this makes sense. Would you be able to take on the PR? Thanks!

travisgroth:
Closed by #38

PR #38, opened 2019-12-30
Allow service and deployment specific annotations

derrickburns:
Addresses #35


travisgroth:
Hi @derrickburns - I think the logic to have per-service overrides is easier to follow if you do something like:

```
  annotations:
{{- if .Values.authenticate.deployment.annotations }}
  {{- range $key, $value := .Values.authenticate.deployment.annotations }}
    {{ $key }}: {{ $value | quote }}
  {{- end }}
{{- else if .Values.annotations }}
  {{- range $key, $value := .Values.annotations }}
    {{ $key }}: {{ $value | quote }}
  {{- end }}
{{- end }}
``` 

You can do a similar approach with `service.annotations`.  This is a slightly more intrusive, but cleaner end state.

Please also be sure to sign the CLA, version bump the chart, and update the README as well.  Heads up - we mirror our chart in `helm/stable`, and there is probably an inbound merge to rebase on shortly.

derrickburns:
I have tried to execute the CLA several times.  Giving up.

desimone:
Sorry for the frustration. 

I’m fine with DCO. I’m going to turn off CLA for all but main repo. Thanks for the PR

travisgroth:
Hey @derrickburns thanks so much.  We had another merge in upstream two days ago but we didn't bring it into this repo yet.  Would you mind rebasing onto it if I merged it into master here?  I think README will have an annoying conflict but otherwise it should be easy. There are two options:
```
| `forwardAuth.nameOverride`          | External name of the forward-auth endpoint                                                                                                                                                                                                                                                           | `forwardauth.${rootDomain}`                                                           |
| `forwardAuth.enabled`               | Enable forward-auth endpoint for third party ingress controllers to use for auth checks.  Setting this disables automatic enumeration of `from` hostnames in the Pomerium Ingress object to prevent conflicts.  Use `ingress.hosts` to mix forward-auth and proxy mode on a single Pomerium instance | `false`                                                                               |
```

travisgroth:
Also LGTM otherwise.

derrickburns:
@travisgroth Sure, however I am not clear what you need me to do.  When I pull, I don't see any changes to master.

values.yaml
```diff
@@ -38,6 +38,8 @@ authenticate:
     defaultSANList: []
     defaultIPList: []
   replicaCount: 1
+  service: {}
+  deploymnet: {}
```
travisgroth: Typo here.

Also, to pass lint:

```
  service: 
    annotations: {}
  deployment: 
    annotations: {}
```

values.yaml
```diff
@@ -50,6 +52,8 @@ authorize:
...
```
travisgroth: See previous comment on linting.

values.yaml
```diff
@@ -65,6 +69,8 @@ proxy:
   authenticateServiceUrl: ""
   authorizeServiceUrl: ""
   authorizeInternalUrl: ""
+  service: {}
+  deploymnet: {}
```
travisgroth: See previous comment on linting.

values.yaml
```diff
@@ -38,6 +38,10 @@ authenticate:
     defaultSANList: []
     defaultIPList: []
   replicaCount: 1
+  service:
```
travisgroth:
```suggestion
  service:
```

values.yaml
```diff
@@ -38,6 +38,10 @@ authenticate:
     defaultSANList: []
     defaultIPList: []
   replicaCount: 1
+  service: 
+    annotations: {}
+  deployment:
```
travisgroth:
```suggestion
  deployment:
```

values.yaml
```diff
@@ -50,6 +54,10 @@ authorize:
     defaultSANList: []
     defaultIPList: []
   replicaCount: 1
+  service:
```
travisgroth:
```suggestion
  service:
```

values.yaml
```diff
@@ -50,6 +54,10 @@ authorize:
     defaultSANList: []
     defaultIPList: []
   replicaCount: 1
+  service: 
+    annotations: {}
+  deployment:
```
travisgroth:
```suggestion
  deployment:
```

values.yaml
```diff
@@ -65,6 +73,10 @@ proxy:
   authenticateServiceUrl: ""
   authorizeServiceUrl: ""
   authorizeInternalUrl: ""
+  service:
```
travisgroth:
```suggestion
  service:
```

values.yaml
```diff
@@ -65,6 +73,10 @@ proxy:
   authenticateServiceUrl: ""
   authorizeServiceUrl: ""
   authorizeInternalUrl: ""
+  service: 
+    annotations: {}
+  deployment:
```
travisgroth:
```suggestion
  deployment:
```

PR #48, opened 2020-01-15
Fixes the annotations in deployments and services

victornoel:
Hi,

Sorry I missed it during the review of the previous PR on this, but there was two bugs with it:
- `annotations` was at the wrong level for services
- when `annotations` wasn't set in the values, a field without any value was generated which I think is illegal.


travisgroth:
>when annotations wasn't set in the values, a field without any value was generated which I think is illegal.
I _think_ that's okay and/or helm ignores it, but no harm in the change to make it explicit.

PR #53, opened 2020-01-31
Set hook deletion policy on TLS secrets

### LINKED ISSUE #51: forceGenerateTLS does not work for an upgrade
When doing an upgrade, if I set the `config.forceGenerateTLS` value to `true`, I get the following error:
```
Error: UPGRADE FAILED: secrets "pomerium-authenticate-tls" already exists
```

I suppose a workaround is to delete manually the secret beforehand, but we should:
- at least document it
- at best provide a safer upgrade path

I'm not exactly clear how to do that though.

I think the problem is maybe that when `config.forceGenerateTLS` is set to `false`, then maybe the resource should be generated so that helm continue to keep track of it? I'm not sure…

travisgroth:
Fixes #51 with short term solution.

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 5.0.1
+version: 5.0.2
 appVersion: 0.6.0
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo-long.svg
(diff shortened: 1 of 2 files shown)


travisgroth:
/CC @victornoel - I tried this quickly this morning.  If you have time to verify it behaves as expected in your environment, that'd be great.

victornoel:
@travisgroth great and simple, I tried to set `forceGenerateTLS` to `true` and the certificates were regenerated without any problem even though they were already present.

So this neatly solve the main problem during upgrade :)

PR #56, opened 2020-02-08
Idp scopes

psychomelet:


psychomelet:
the ci somehow failed with `Error linting charts: Error processing charts`

I don't quit understand that since I only changed 3 lines.

desimone:
@psychomelet thank you for the PR. 

The tests failed because of the requirement to bump the chart. 

```
 ✖︎ pomerium => (version: "5.0.3", path: "charts/pomerium") > Chart version not ok. Needs a version bump!
```

Which is a holdover from test suite used on `helm/stable`.

psychomelet:
also changed the policy value from dict to list, to avoid the warning
```
coalesce.go:196: warning: cannot overwrite table with non table for policy (map[])
```

travisgroth:
Thanks @psychomelet.  Can you squash and sign the commits (see https://probot.github.io/apps/dco/).  LGTM otherwise.

psychomelet:
all fixed

Issue #57, opened 2020-02-08
Needs more documentation on behaviors for new users


psychomelet:
Encountered some problems when using values.yaml with configmap, I'll list some problems.

I followed the doc with 2 steps:
1. created config.yaml and add it to configmap
2. helm install with existingConfig set to configmap value.

Now the problems:

sometimes values.yaml config overwrite configmap
1. authenticate.idp settings was in deployment env and  overwrite the config.yaml values. So we can't put the idp config inside config.yaml.
2. no scopes support in authenticate.idp which I submit an PR for it, while AWS Cognito need idp.scopes to be set. With 1 in mind, you can't set it in the configmap, cause if IDP_* found in the environment, the idp_scopes settings in configmap.yaml somehow ignored.
3. same with shared_secret and cookie_secret

sometimes the values.yaml config won't overwrite configmap
1. policy settings configured in values.yaml will not overwrite configmap when existingConfig set, it does nothing.
2. but if we add policy in the config.yaml, then the ingress route will not be configured.

tweaks essential but commented out... these really needs documentation:
1. I find `nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"` is essential to my k8s deployment on Aliyun Cloud, have not tested other. Typically cloud provider terminates the https after load balancer, it's essential to be enabled.
2. The `nginx.ingress.kubernetes.io/proxy-buffer-size: "16k"` is essential to my deployment, have not tested other. The Azure AD callback url is too large for default size. If the setting not enabled, it just hang there.

travisgroth:
Hi @psychomelet - Thanks for trying pomerium.  You've got a few things going on here.  

I do want to start off suggesting using the chart-generated config rather than using an existing config.  As you've seen, it tends to work a lot more smoothly.  That's the path most people use and it is the most tested and thought about.  That said, it should work in general.

>authenticate.idp settings was in deployment env and overwrite the config.yaml values. So we can't put the idp config inside config.yaml.

Correct.  This is a bit of legacy configuration showing.  Folding those env vars into the main config is very doable but we'll need to make sure the checksumming picks them up correctly.  Those values aren't dynamic and having them as part of the pod spec forces updates automatically.

>no scopes support in authenticate.idp

Yes that appears to be a miss.  Happy to accept that PR.

>cause if IDP_* found in the environment, the idp_scopes settings in configmap.yaml somehow ignored.

[...]

PR #58, opened 2020-02-12
Update README

travisgroth:
Address some comments in #57


.github/workflows/test.yaml
```diff
@@ -26,6 +26,8 @@ jobs:
       - name: Install helm
         uses: azure/setup-helm@v1
         id: install
+        with:
+          version: v3.0.3
```
desimone: 🙄

travisgroth: 💩

PR #60, opened 2020-02-15
fix(helpers): invalid reference to existingCASecret

hugocortes:
When providing an existing CA secret by `config.existingCASecret`, the template is incorrectly referencing `.Values.proxy.existingCASecret` to be used for `pomerium.caSecret.name` template.

error returned:
```
Error: template: pomerium/templates/_helpers.tpl:232:43: executing "pomerium.caSecret.name" at <63>: invalid value; expected string
helm.go:76: [debug] template: pomerium/templates/_helpers.tpl:232:43: executing "pomerium.caSecret.name" at <63>: invalid value; expected string
```

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 5.0.5
+version: 5.0.6
 appVersion: 0.6.2
 home: http://www.pomerium.io/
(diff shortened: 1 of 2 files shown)


desimone:
https://github.com/pomerium/pomerium-helm/blob/9973a3d2dd1431db788bbcc75f8e9bff26c631a4/charts/pomerium/values.yaml#L14

👀

PR #61, opened 2020-02-17
v6.0.0 - Add experimental pomerium-operator support, deprecate legacy TLS Secret support

travisgroth:


charts/pomerium/README.md
```diff
@@ -57,6 +58,16 @@ helm delete --purge my-release
 
 The command removes nearly all the Kubernetes components associated with the chart and deletes the release.
 
+## Pomerium Operator (EXPERIMENTAL)
```
desimone: 🔥

charts/pomerium/templates/authorize-deployment.yaml
```diff
@@ -54,10 +56,9 @@ spec:
         image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         args:
-{{- if or .Values.config.existingConfig .Values.config.policy }}
+        {{- if or (or .Values.config.existingConfig .Values.config.policy) .Values.operator.enabled }}	
           - --config=/etc/pomerium/config.yaml
-{{- end }}
-{{- range $key, $value := .Values.extraArgs }}
+        {{- end }}{{- range $key, $value := .Values.extraArgs }}
```
desimone: nit:  I think the for loop pops out a bit more on a seperate line

charts/pomerium/templates/configmap.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if not .Values.config.existingConfig }}
+{{- if .Values.operator.enabled }}
```
desimone: This looks off. We'd lose these annotations in non operator mode right?

travisgroth: The logic here is a little odd.  In operator mode we have two config maps - one empty, and one populated from the static values of the chart.  When running without the operator, we only need the main config map with static values.

To save a big repeated block in the conditional, I am changing the name of the ConfigMap with the static values depending on whether we're in operator mode.  Similarly, I just flip the empty ConfigMap off in non-operator mode.

Hopefully that tracks what's in the yaml...unless I'm missing something else.

charts/pomerium/templates/authorize-deployment.yaml
travisgroth: Good catch.

charts/pomerium/README.md
```diff
@@ -157,9 +168,24 @@ A full listing of Pomerium's configuration variables can be found on the [config
...
+| `operator.config.debug`               | Enable Pomerium Operator debug logging                                                                                                                                                                                                                                                             | `false`                                                                               |
+| `operator.deployment.annotations`     | Annotations for the operator deployment.                                                                                                                                                                                                                                                           | `{}`                                                                                  |
 
 ## Changelog
 
+### 6.0.0
+
+- Integrate pomerium operator
+- Remove legacy TLS config support.  See [v3.0.0 Upgrade Notes](#300-1) to migrate
```
desimone: 👍

charts/pomerium/templates/configmap.yaml
desimone: Makes  complete sense. 👍

PR #66, opened 2020-03-03
Signing key in secret

### LINKED ISSUE #65: Setting SIGNING_KEY in the secret
Hello,
we use `SIGNING_KEY` setting to enable the JWT token in response header in order to get additional user informations that are not available in other headers.

The chart has no parameter for that `SIGNING_KEY`, therefore we have to give it through `extraEnv` parameter which is not secure since it is not encoded in our configuration file.

It would be great to add a `proxy.signingKey` parameter to the chart and store it in the secret, allowing to give it through an existing secret and therefore not expose it in the configuration file.

tgomas:
Here is a proposition to add a `proxy.signingKey` parameter which is stored in the secret.

close #65


charts/pomerium/templates/secret.yaml
```diff
@@ -13,6 +13,7 @@ type: Opaque
 data:
   cookie-secret: {{ default (randAscii 32 | b64enc | b64enc) (.Values.config.cookieSecret | b64enc )}}
   shared-secret: {{ default (randAscii 32 | b64enc | b64enc) (.Values.config.sharedSecret | b64enc )}}
+  signing-key: {{ .Values.proxy.signingKey | b64enc }}
```
victornoel: why not generate it like the rest above?

tgomas: Generating it would change default behavior. By default the value is an empty string not enabling JWT token in response headers.

travisgroth: It'd be nice to have it enabled by default - I don't think that would break anything.  

To do that correctly, I think we'd move it into a TLS secret with hook annotations like the https certs.  I believe  - @desimone can you confirm - pomerium will be implicitly doing this anyway in the next version, so we might as well try to do the right thing out of the box.

@tgomas would you mind taking a stab at that?  I can always do a follow up, otherwise.

desimone: @travisgroth That's right. If no signing key is provided, the authorize service (moved from proxy) generates one and prints out the public key.

https://github.com/pomerium/pomerium/pull/514/files#diff-d1e68c602dfe07e82a613b34bafa753bR54-R76

tgomas: Thank you for explanations and advice about hook annotation.
@travisgroth I'll look further on that way

charts/pomerium/README.md
```diff
@@ -135,6 +135,8 @@ A full listing of Pomerium's configuration variables can be found on the [config
...
+| `proxy.existingJWTSecret`             | Name of existing JWT Secret for proxy requests.                                                                                                                                                                                                                              |                                                                                       |
+| `proxy.jwt.signingKey`                    | Signing key is the base64 encoded key used to sign outbound requests.                                                                                                                                                                                                                              |                                                                                       |
```
victornoel: @tgomas the spacing is broken

charts/pomerium/templates/_helpers.tpl
```diff
@@ -210,6 +210,20 @@ Adapted from : https://github.com/helm/charts/blob/master/stable/drone/templates
 {{- end -}}
 {{- end -}}
 
+{{/* Determine secret name for Proxy JWT signing key */}}
+{{- define "pomerium.proxy.jwtSecret.name" -}}
```
victornoel: @tgomas I wonder if it wouldn't be best to keep the same naming as the variable and the configuration, i.e., `signingKey` without using the terms `jwt` for example. `pomerium.signingKeySecret.name`.

Same for the the values: `proxy.signingKey` and `proxy.existingSigningKeySecret` seem good.

victornoel: (and same for the names below, jwt shouldn't appear in the name IMHO)

charts/pomerium/templates/proxy-deployment.yaml
```diff
@@ -79,6 +79,11 @@ spec:
             secretKeyRef:
               name: {{ $secretName }}
               key: shared-secret
+        - name: SIGNING_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "pomerium.proxy.jwtSecret.name" . }}
+              key: signing.key
```
victornoel: @tgomas why use `.` as a separator, other secrets seem to use a `-`.

charts/pomerium/templates/proxy-deployment.yaml
```diff
@@ -79,6 +79,11 @@ spec:
             secretKeyRef:
               name: {{ $secretName }}
               key: shared-secret
+        - name: SIGNING_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "pomerium.proxy.jwtSecret.name" . }}
```
victornoel: @tgomas why use another secret and not the same used currently for secrets, see for example `COOKIE_SECRET` and `SHARED_SECRET` above, I don't see a reason to introduce a new secret just for the signing key, is there?

victornoel: signing key seems to be at the same abstraction level than cookie secret and shared secret IMHO

charts/pomerium/templates/proxy-deployment.yaml
victornoel: actually, it's not so obvious, see the certificates secret that use a `.`, @travisgroth an opinion?

charts/pomerium/templates/proxy-deployment.yaml
victornoel: actually, maybe it could also be argued that they are at the same level than the certificates for proxy… @travisgroth @desimone an opinion?

charts/pomerium/templates/secret.yaml
tgomas: Now the signing key is managed as TLS secrets, that means it is generated at install and can be forced to be regenerated like TLS secrets.

However I wonder that having this JWT signing key secret being defined in `tls-secrets.yaml` and its generation being activated by `generateTLS` and `forceGenerateTLS` could be misleading. On the other hand, adding JWT dedicated file and values to manage only one signing key is probably not what we want since it make sense to (re)generate all those secrets at same time.

@travisgroth have you any idea on how to deal with that ?

charts/pomerium/templates/tls-secrets.yaml
```diff
@@ -187,4 +213,9 @@ data:
   {{ template "pomerium.proxy.tlsSecret.certName" . }}: {{ .Values.proxy.tls.cert | b64enc }}
   {{ template "pomerium.proxy.tlsSecret.keyName" . }}: {{ .Values.proxy.tls.key | b64enc }}
 {{-     end }}
+{{-     if and (not .Values.proxy.existingJWTSecret) .Values.proxy.jwt.signingKey }}
+---
+{{ template "pomerium.proxy.jwtSecretObject" . }}
+  {{ template "pomerium.proxy.jwtSecret.keyName" . }}: {{ .Values.proxy.jwt.signingKey | b64enc }}
```
tgomas: This keyName is named to be consistent with certName and keyName from TLS secrets, but none seems to be defined anywhere. So I suppose I'm missing something.

@travisgroth could you help me to figure out where it should be defined ?

charts/pomerium/templates/proxy-deployment.yaml
tgomas: because TLS secrets use a `.` (e.g. `tls.key` or `ca.crt`)

charts/pomerium/templates/proxy-deployment.yaml
tgomas: To no more manage it with helm after it has been generated

It could be stored in `pomerium.proxy.tlsSecret.name` since they have same lifecycle, but it would be misleading since JWT signing key is not related to TLS. That's close to the point discussed here: https://github.com/pomerium/pomerium-helm/pull/66#discussion_r393092226

charts/pomerium/templates/_helpers.tpl
tgomas: I introduced `jwt` in naming to highlight the difference with `tls` secrets since `signingKey` would not be clear in a context where everything is related to TLS.

It could make sense to extract secret's definition in `signing-key-secret.yaml` and rename everything with `signingKey` instead `jwt`. In that case, should the secret also require dedicated values to switch its generation on since it will no more be related to tls secrets ?

victornoel: I think the fact the generated signing key is a gpg key is a different matter from the fact that a tls secret (which is a special kind of secret in k8s) is used.

IMHO:
- the signing key should be in a normal secret resource
- the signing key should be generated in the same manner as tls secrets resource
- the signing key could/should be stored with cookie and shared secrets in the same secret resource

tgomas: @travisgroth what do you think about that ?

It's different from [your first advice](https://github.com/pomerium/pomerium-helm/pull/66#discussion_r389578953) but I can make changes in that way.

travisgroth: re: naming, I'm partial to sticking close to the parameter name in pomerium which is just `SIGNING_KEY`, for better or worse.  No hard opinion here though.  

>the signing key should be in a normal secret resource

I was actually thinking a TLS secret.  Isn't it just another x509 keypair?  Operationally it is unimportant either way, though.  

>the signing key should be generated in the same manner as tls secrets resource

Yes

>the signing key could/should be stored with cookie and shared secrets in the same secret resource

This won't work with hooks, unfortunately.  It will need to be its own object to be controlled correctly.

charts/pomerium/templates/proxy-deployment.yaml
travisgroth: See other discussion on this one

charts/pomerium/templates/tls-secrets.yaml
```diff
@@ -105,6 +105,28 @@ type: kubernetes.io/tls
...
+    app.kubernetes.io/name: {{ template "pomerium.name" . }}
+    helm.sh/chart: {{ template "pomerium.chart" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    helm.sh/hook-delete-policy: before-hook-creation
+{{- if .Values.config.forceGenerateTLS }}
+    helm.sh/hook: pre-upgrade
+{{- else if .Values.config.generateTLS }}
+    helm.sh/hook: pre-install
```
travisgroth: I _think_ we should have an independent flag for the signing key generation.

charts/pomerium/templates/tls-secrets.yaml
travisgroth: ...oh I think you found a bug.  Those were previously defined in `_helpers.tpl` to support a legacy secret layout, which was removed in 6.0.0.  It looks like these structures weren't updated.

They are now hard coded since they're proper TLS secrets like so - https://github.com/pomerium/pomerium-helm/blob/9ea4dbcbde8533eaf44e0b4c21d5a619dba5b568/charts/pomerium/templates/tls-secrets.yaml#L166-L167

So these should be just set to the field names we settle on in the other discussion.

charts/pomerium/README.md
```diff
@@ -91,11 +91,32 @@ You may force recreation of your TLS certificates by setting `config.forceGenera
...
+## Signing Key
+
+### Auto Generation
+
+In default configuration, this chart will automatically generate a signing key in a helm `pre-install` hook for the Pomerium proxy to sign jwt sent in responses.
+
+Upon delete, you will need to manually delete the generated secret. Example:
+
+```console
+kubectl delete secret -l app.kubernetes.io/name=pomerium
```
victornoel: this is going to delete all secret with this label, is that the objective?

charts/pomerium/README.md
```diff
@@ -91,11 +91,32 @@ You may force recreation of your TLS certificates by setting `config.forceGenera
...
+```
+
+You may force recreation of your signing key by setting `config.forceGenerateSigningKey` to `true`. Delete already existing signing key secret first to prevent errors, and make sure you set back to `false` for your next helm upgrade command or your deployment will fail due to existing Secret.
+
+### Self Provisioned
+
+If you wish to provide your own signing key in secret, you should:
+
+1. turn `config.generateSigningKey` to `false`
```
victornoel: @tgomas this option does not seem to be documented below

charts/pomerium/README.md
```diff
@@ -135,6 +156,8 @@ A full listing of Pomerium's configuration variables can be found on the [config
...
+| `proxy.existingSigningKeySecret`      | Name of existing Signing key Secret for proxy requests.                                                                                                                                                                                                                              |                                                                                       |
+| `proxy.signingKey`                    | Signing key is the base64 encoded key used to sign outbound requests.                                                                                                                                                                                                                              |                                                                                       |
```
victornoel: @tgomas is this still used?

charts/pomerium/README.md
victornoel: also `forceGenerateSigningKey`

charts/pomerium/templates/proxy-deployment.yaml
```diff
@@ -79,6 +79,11 @@ spec:
             secretKeyRef:
               name: {{ $secretName }}
               key: shared-secret
+        - name: SIGNING_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "pomerium.proxy.signingKeySecret.name" . }}
+              key: signing-key
```
victornoel: @tgomas I don't see the code that generates the secret anymore

charts/pomerium/templates/signing-key-secret.yaml
```diff
@@ -0,0 +1,32 @@
...
+{{- if .Values.config.forceGenerateSigningKey }}
+    helm.sh/hook: pre-upgrade
+{{- else if .Values.config.generateSigningKey }}
+    helm.sh/hook: pre-install
+{{- end }}
+  name: {{ template "pomerium.proxy.signingKeySecret.name" . }}
+  namespace: {{ .Release.Namespace }}
+type: Opaque
+data:
+{{- end -}}
```
victornoel: @tgomas is this end at the right place? there seem to be something not right with the key being generated just after

charts/pomerium/README.md
tgomas: Not really, we can have a more specific example like:

`kubectl delete secret -l name=pomerium-xxxxx-signing-key`

But the problem is the same with same example for deleting tls secrets. Should it be also more specific with example for each tls secret ?

charts/pomerium/README.md
tgomas: Yes, like for tls secrets it is still possible to directly set the value

charts/pomerium/templates/proxy-deployment.yaml
tgomas: it's in `signing-key-secret.yaml`

charts/pomerium/README.md
victornoel: ok, I see there was a missing file when I made the comment :)

charts/pomerium/templates/proxy-deployment.yaml
victornoel: this seems to have been fixed

charts/pomerium/README.md
victornoel: actually to delete a secret by name, use `kubectl delete secret pomerium-xxxxx-signing-key`.

> But the problem is the same with same example for deleting tls secrets. Should it be also more specific with example for each tls secret ?

I don't know, actually the example for the TLS secret is not correct anymore now that we have multiple secrets with this label…

charts/pomerium/templates/signing-key-secret.yaml
tgomas: yes it's the `end` of `pomerium.proxy.signingKeySecretObject`'s definition, which is used in both if and else blocs below. The key is only generated inside if bloc because not used in else bloc.

victornoel: ah it's not an if at the top, sorry

charts/pomerium/README.md
tgomas: Basic secret (with `cookie-secret` and `shared-secret`) is also deleted by the example for TLS secret.

Should we add another annotation with secret type in order to show more selective examples ?

charts/pomerium/README.md
tgomas: done

charts/pomerium/templates/signing-key-secret.yaml
tgomas: While fixing key generation (it has to be b64enc 2 times) I removed the key variable declaration since it was used at only one place. Hope that's more readable

charts/pomerium/templates/_helpers.tpl
tgomas: Signing key is only a key and not a keypair. Therefore it is now store in dedicated Opaque secret

charts/pomerium/templates/proxy-deployment.yaml
tgomas: now that it's in an Opaque secret, a `-` is used as separator

charts/pomerium/README.md
travisgroth: We should probably update the docs to just reflect the names.  More labels sounds like overkill to me.

tgomas: done

charts/pomerium/templates/tls-secrets.yaml
tgomas: done

charts/pomerium/README.md
tgomas: done

charts/pomerium/README.md
```diff
@@ -82,7 +82,11 @@ In default configuration, this chart will automatically generate TLS certificate
 Upon delete, you will need to manually delete the generated secrets. Example:
 
 ```console
-kubectl delete secret -l app.kubernetes.io/name=pomerium
+kubectl delete secret -l name=pomerium-authenticate-tls
```
victornoel: @tgomas you don't need the `-l name=`, `kubectl delete secret pomerium-authenticate-tls` is enough.

charts/pomerium/templates/signing-key-secret.yaml
```diff
@@ -0,0 +1,31 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    helm.sh/hook-delete-policy: before-hook-creation
+{{- if .Values.config.forceGenerateSigningKey }}
+    helm.sh/hook: pre-upgrade
+{{- else if .Values.config.generateSigningKey }}
+    helm.sh/hook: pre-install
+{{- end }}
+  name: {{ template "pomerium.proxy.signingKeySecret.name" . }}
+  namespace: {{ .Release.Namespace }}
```
victornoel: @tgomas helm already takes care of setting the namespace, it's bad practice to include it in templates

tgomas: @victornoel both basic and tls secrets templates have same namespace definition, I wonder why ?

Also do you recommend to remove namespace definition in this PR from all secrets templates, or only from signing-key-secret.yaml and optionally from other secrets in a separate PR ?

victornoel: @tgomas it's a mistake IMHO, let's remove all of them if @travisgroth is ok, if not just for this one then :)

charts/pomerium/README.md
tgomas: done

charts/pomerium/templates/signing-key-secret.yaml
tgomas: Removed for all secrets template, can be changed if @travisgroth has another point of view

travisgroth: Likely just been copied and pasted.  👍 on removing them.

tgomas:
@travisgroth all comments have been processed: signing-key can be generated at install and is stored in a dedicated secret. Could you please help me to get it merged ? I'm still open to improve it if you have any additional comment.

travisgroth:
@tgomas FYI, I had to unpublish 6.0.5 as this does actually cause a breaking change to existing charts.  I'm going to re-release as 7.0.0.  That's my mistake for not catching it.  

I will also add instructions similar to what we had for the 5.x release https://github.com/pomerium/pomerium-helm/tree/master/charts/pomerium#500-1.

PR #70, opened 2020-03-21
Feature/store config in secret

### LINKED ISSUE #67: Provide the possibility to store sensitive headers in secrets
Hello.

`policy` allows to set custom headers via the `set_request_headers` section.
Headers can store sensitive data like bearer token for the Kubernetes dashboard.

The chart stores the policy in the `configmap`. k8s `secret` looks like a safer option.

It would be great if we could store `policy` or just sensitive bits of a `policy` in a `secret`.

I can a PR if you don't mind.

Thanks in advance.

### LINKED ISSUE #71: Chart doesn't support operator with existing config
I'm trying to use the chart with the following configuration:

```yaml
config:
  existingConfig: "config"
operator:
  enabled: true
```

Seems that there are 2 problems:
1. existing config is mounted to the worker (authorize/proxy/etc.) pods instead of the new one generated by the operator
2. new empty config is not created during the chart installation

abuzhynsky:
Fixes #67.
Fixes #71.

I've done my best with the naming, hope that config stored in the `Secret` is not confusing.

I haven't added automatic migration for the `existingConfig` option - I don't see an easy solution here.

I've used `stringData` for the new `Secret` - I think that it's a bit easier to read than the named template and conversion to `base64` but I can change it for consistency with the existing secrets from the chart.

I've bumped the chart version (not sure if I should do it).

`operator.image.tag` also should be changed after [the PR in the operator](https://github.com/pomerium/pomerium-operator/pull/44) will be merged.

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 7.0.1
+version: 8.0.0
 appVersion: 0.6.2
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo-long.svg
--- a/charts/pomerium/README.md
+++ b/charts/pomerium/README.md
@@ -16,6 +16,7 @@
     - [Self Provisioned](#self-provisioned-1)
   - [Configuration](#configuration)
   - [Changelog](#changelog)
+    - [8.0.0](#800)
     - [7.0.0](#700)
     - [6.0.0](#600)
     - [5.0.0](#500)
@@ -23,6 +24,7 @@
     - [3.0.0](#300)
     - [2.0.0](#200)
   - [Upgrading](#upgrading)
+    - [8.0.0](#800-1)
     - [7.0.0](#700-1)
     - [5.0.0](#500-1)
     - [4.0.0](#400-1)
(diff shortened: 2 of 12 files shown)


travisgroth:
Thanks abuzhynsky!  I'm going to start with reviewing the operator changes (since that really has to be ready for this PR) and try to get a few smaller 6.x changes in before we merge this.

travisgroth:
> I haven't added automatic migration for the existingConfig option - I don't see an easy solution here.

There's a section for upgrade notes.  I think a one-liner that does the migration would be fine.  I believe we've also put scripts into the repo in the past for more complicated upgrades.
 
> I've used stringData for the new Secret - I think that it's a bit easier to read than the named template and conversion to base64 but I can change it for consistency with the existing secrets from the chart.

One concern - I haven't used stringData but does helm-diff handle it correctly?  While not everyone uses the diff tool in their environment, I feel it is common enough to make sure we support well enough.  

> I've bumped the chart version (not sure if I should do it).

Yes, this is definitely a major semver change.

> operator.image.tag also should be changed after the PR in the operator will be merged.

Let's have that tag as part of this PR; we just need to work through the operator release first.

abuzhynsky:
> does helm-diff handle it correctly?

I've tested it - [`helm-diff`](https://github.com/databus23/helm-diff) is fine with the `stringData` and shows diff only if it was changed.

travisgroth:
> I've tested it - [`helm-diff`](https://github.com/databus23/helm-diff) is fine with the `stringData` and shows diff only if it was changed.

OK great.  Let's stick with stringData then.

abuzhynsky:
@travisgroth 
I've updated the PR, please take a look.

I've consolidated 2 secrets (for config and the existing one) into the single secret, removed ENVs that were mounted from Secret and added the fix for #71.

The only thing is I haven't added the automatic migration for `existingConfig` and `existingSecret`. Looks like it requires some `jq` magic and even with it can't be done automatically if only one part (`secret` or `config`) was used.

abuzhynsky:
@travisgroth 
I've rebased and squashed my changes.
Looks I'm late for the `7.0.0` release so I've bumped the version to the `8.0.0`.
Please take a look again.

travisgroth:
>Looks I'm late for the 7.0.0 release

That was planned to be a patch release but there were some breaking changes so we had to re-release as 7.

Can you rebase the operator changes so we can bump the referenced image version in this PR?  It needs to be all in one shot or we'll have a broken version.  Thanks!

abuzhynsky:
@travisgroth 
I've updated the operator image tag and rebased my changes onto the latest master.
Please take a look.

PR #73, opened 2020-03-23
Fix tls secrets keynames

### LINKED ISSUE #72: Missing key names for TLS secrets provided in values yaml file
They are now hard coded since they're proper TLS secrets like so - https://github.com/pomerium/pomerium-helm/blob/9ea4dbcbde8533eaf44e0b4c21d5a619dba5b568/charts/pomerium/templates/tls-secrets.yaml#L166-L167

So these should be just set to the field names we settle on in the other discussion.

_Originally posted by @travisgroth in https://github.com/pomerium/pomerium-helm/pull/66_

tgomas:
close #72 

tls secrets keynames are hardcoded
also key and cert values can be specified for ca secret

### DIFF
--- a/charts/pomerium/templates/tls-secrets.yaml
+++ b/charts/pomerium/templates/tls-secrets.yaml
@@ -159,32 +159,33 @@ data:
   tls.crt: {{ $kp.Cert | b64enc }}
   tls.key: {{ $kp.Key | b64enc }}
 {{-   else if not .Values.config.generateTLS }}
-{{-     if and (not .Values.config.existingCASecret) .Values.config.ca }}
+{{-     if and (not .Values.config.existingCASecret) .Values.config.ca.cert .Values.config.ca.key }}
 {{ template "pomerium.ca.tlsSecretObject" . }}
-  {{ template "pomerium.caSecret.certName" . }}: {{ .Values.config.ca | b64enc }}
+  ca.crt: {{ .Values.config.ca.cert | b64enc }}
+  ca.key: {{ .Values.config.ca.key | b64enc }}
 {{-     end }}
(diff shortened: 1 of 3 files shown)


tgomas:
@travisgroth could you please review this fix about the bug found while working on [signing-key secret](https://github.com/pomerium/pomerium-helm/pull/66#discussion_r393399513)

charts/pomerium/templates/tls-secrets.yaml
```diff
@@ -159,32 +159,33 @@ data:
...
+  ca.crt: {{ .Values.config.ca.cert | b64enc }}
+  ca.key: {{ .Values.config.ca.key | b64enc }}
 {{-     end }}
 {{-     if and (not .Values.authenticate.existingTLSSecret) .Values.authenticate.tls.cert .Values.authenticate.tls.key }}
 ---
 {{ template "pomerium.authenticate.tlsSecretObject" . }}
-  {{ template "pomerium.authenticate.tlsSecret.certName" . }}: {{ .Values.authenticate.tls.cert | b64enc }}
-  {{ template "pomerium.authenticate.tlsSecret.keyName" . }}: {{ .Values.authenticate.tls.key | b64enc }}
+  tls.crt: {{ .Values.authenticate.tls.cert | b64enc }}
+  tls.key: {{ .Values.authenticate.tls.key | b64enc }}
```
victornoel: @tgomas `pomerium.authenticate.tlsSecret.keyName` and others shouldn't have been removed from templates then?

tgomas: They were not in templates, hence the bugfix

victornoel: ok, thx for the explanation :)

PR #74, opened 2020-03-23
Add support for administrators to be provided via helm values

shreyu86:
* Have added support for passing administrators to the deployments created via. helm

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 6.0.3
+version: 6.0.4
 appVersion: 0.6.2
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo-long.svg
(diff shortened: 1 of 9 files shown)


travisgroth:
Thanks shreyu86 - I think it would be cleaner and simpler to add this to the config.yaml, rather than as a discrete key in the pomerium secret.  The authorize service should pick this up dynamically.

Also worth noting that you could pass this in as a `config.extraOpts`, which seems to not be documented.  I'm happy to have this as a first class option, though.

shreyu86:
Thanks for the feedback @travisgroth I did not know that administrators could be added to config.yaml I agree it would be much cleaner that way. Happy to remove the secret. As a clarification question it would be sufficient to add administrators to config block in values.yaml is that correct?

travisgroth:
> As a clarification question it would be sufficient to add administrators to config block in values.yaml is that correct?

You'll need to define it there to pass linting but also render it explicitly in the configmap template.

shreyu86:
@travisgroth changed based on your feedback also sorry, I did forget adding https://github.com/pomerium/pomerium-helm/pull/74/files#diff-346b22ac270037b05f4899e2d423e281R23 in the first round of changes.

PR #76, opened 2020-03-29
Add helm hooks to operator-managed configmap

rocketspacer:
Without hooks, the operator-managed ConfigMap get nullified every apply.
This causes brief windows of outage, until the operator syncs the data back.

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 7.0.0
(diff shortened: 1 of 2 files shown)


travisgroth:
@rocketspacer are you deploying with helm 2 or helm 3?  That sounds like a helm 3 behavior with the new merge strategy but maybe I'm mistaken.

I'm wondering if omitting the `data` field or the `config.yaml` key would keep it from being considered.  While a hook is valid, I'm becoming a bit conscious of how much we're relying on them and would love to avoid if possible.

PR #77, opened 2020-03-30
Fix cache service annotations indentation

victornoel:
Also add the same test in front of it as other services

I think I found it because helm 3 refuses such an error while helm 2 didn't, maybe we should add helm 3 tests too?


travisgroth:
>maybe we should add helm 3 tests too?

It looks like the chart testing action has moved up to helm 3, so that should be possible now.  Opened https://github.com/pomerium/pomerium-helm/issues/79 for that.

PR #83, opened 2020-04-19
Adding insecure server boolean to chart.

bjoernw:
The chart was lacking support for a way to run pomerium in insecure mode. My use case is that I would like to fold pomerium into my istio mesh and since istio already handles mutual-tls I did not need/want pomerium to do that as well. Since pomerium itself already supports an insecure mode I just needed to expose this toggle in the chart.

However, just setting environment variables INSECURE_SERVER and GRPC_INSECURE wasn't enough to get past the failing liveness and readiness checks which still had a scheme of "HTTPS" regardless of the value of INSECURE_SERVER or GRPC_INSECURE.

Here I create a single flag `config.insecure` that abstracts those details away. 

Let me know what you think.


charts/pomerium/templates/NOTES.txt
```diff
@@ -48,3 +48,11 @@ the official documentation:
 
     https://www.pomerium.io/docs/identity-providers.html
 {{- end }}
+
+{{- if .Values.config.insecure }}
+##############################################################################
+####        DANGER: You have disabled TLS!
+####
+####        Please only do this if you absolutely know what you are doing.
+##############################################################################
```
desimone, line 57: 👍 Love the disclaimer.

desimone:
@bjoernw just needs a version bump

bjoernw:
@desimone What version do you want me to bump this to? 8.3.0? Should I capture my change in the changelog in the readme?

desimone:
>@desimone What version do you want me to bump this to? 8.3.0? 

Seems good to me, going to ping @travisgroth to make sure this is a minor not a major version bump. I don't think it breaks existing so... should be fine. 

> Should I capture my change in the changelog in the readme?

Please do!

bjoernw:
@desimone bumped it to 8.4.0 and added changelog. Looks good?

charts/pomerium/README.md
```diff
@@ -240,6 +242,10 @@ A full listing of Pomerium's configuration variables can be found on the [config
 
 ## Changelog
 
+### 8.0.4
```
desimone, line 245: typo I think... `8.4.0` right?

bjoernw: oops, my bad, fixed

desimone:
Thanks @bjoernw !

PR #90, opened 2020-04-29
add new value, forwardAuth.internal

### LINKED ISSUE #89: Option to not expose the forwardAuth over ingress/for internal cluster use only
There should be an option to not expose forwardAuth over the ingress and only use the service (`pomerium-proxy`).

Background is:
https://github.com/pomerium/pomerium/issues/549#issuecomment-603587366
https://github.com/pomerium/pomerium/issues/616#issuecomment-617812685

If forwardAuth is exposed the config is set to `forward_auth_url: forwardauth.${rootDomain}` which means the service name (`pomerium-proxy.svc`) can't be used in forwardAuth mode. They need to match with the config somehow.

Therefore it should be possible to set the `forward_auth_url` to pomerium-proxy.svc. 
It is not possible to just set forwardAuth.nameOverride=`pomerium-proxy.svc` because then it becomes impossible to issue lets encrypt certs for that domain if ingress is created. 

I think the best option would be to just not create an ingress and make forwardAuth only be used internally.

Vad1mo:
this PR /closes #89


charts/pomerium/templates/secret.yaml
```diff
@@ -56,11 +56,13 @@ stringData:
...
 {{- end -}}
-{{- if .Values.forwardAuth.enabled }}
+{{- if and .Values.forwardAuth.enabled .Values.forwardAuth.internal }}
+    forward_auth_url: https://{{ template "pomerium.proxy.fullname" . }}
+{{ else }}
     forward_auth_url: https://{{ template "pomerium.forwardAuth.name" . }}
 {{- end -}}
 {{- if .Values.config.policy }}
-    policy: 
+    policy:
```
desimone, line 65: accidental change?

Vad1mo: YAML Format on save action

Vad1mo:
Signed my commit

desimone:
thanks @Vad1mo !

PR #94, opened 2020-04-30
Fwdauth internal fqdn

albertrdixon:
## What

- Add template for generating forwardauth internal fqdn (fqdn of the proxy service object)
- Use forwardauth internal fqdn as `forward_auth_url` if `fowardAuth.internal==true`

## Why

Fixes #93 

- The nginx ingress controller may not be able to resolve not fully qualified names set in `auth-url`, so the full `service.ns.svc.cluster.local` name is required.
- The proxy is expecting the exact hostname set in `forward_auth_url` and obviously `service-name != service-name.ns.svc.cluster.local`

PR #100, opened 2020-05-07
Fix: when `forwardAuth.internal == true` set also the namespace 

Vad1mo:
Fix: Set not only the service but also the namespace when `forwardAuth.internal == true`

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 8.5.4
+version: 8.5.5
 appVersion: 0.7.6
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo-long.svg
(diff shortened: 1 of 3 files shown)


travisgroth:
This looks like it's fixing a similar problem as #94.  I think that one is a little more flexible though.  Is that accurate, and would you mind if we went with that implementation?  

Question - has this type of setup been fully tested?  I believe users need to be able to hit the forward-auth url in their browsers for that flow to work.  Are you using an independent ingress with a host header override to the proxy?

Vad1mo:
I am not sure ab out cases where a FQDN like proposed in #94 would be needed. But it seems that the PR #94 addresses the same problem like this fix. 

To avoid multiple switches that exclude each other.  (internal=true ignores name_override) I would rather deduce the internalClusterDomain from the actual cluster. (if possible)

travisgroth:
Okay I think this makes sense.  We can improve upon it if needed.

PR #101, opened 2020-05-12
Authorize signing key environment variable

### LINKED ISSUE #98: Authorize signing key missing
```
Chart: pomerium-8.5.1
AppVersion: 0.7.5
```

I'm using forwardauth with the nginx ingress controller and trying to have my backend service validate the JWT assertion after getting a request authenticated by pomerium. I am using a self generated EC key that we deploy alongside pomerium so I am sure the public key I am using is legit. 

Looking at the authorize deployment, I don't see `SIGNING_KEY` set. If I edit the deployment and add my signing key then all is well as expected.

https://github.com/pomerium/pomerium-helm/blob/a0c125fff3f506e66aa9665e4814e390abd52c9b/charts/pomerium/templates/authorize-deployment.yaml#L68-L86

albertrdixon:
## What

- Add environment variable `SIGNING_KEY` to authorize deployment

## Why

- Fixes #98 
- The signing key used by the proxy and authorize service should match so downstream services can perform header validation to ensure received traffic is legit.


charts/pomerium/templates/authorize-deployment.yaml
```diff
@@ -68,6 +68,11 @@ spec:
         env:
         - name: SERVICES
           value: authorize
+        - name: SIGNING_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "pomerium.proxy.signingKeySecret.name" . }}
```
travisgroth, line 74: This parameter was actually moved from `proxy` to `authorize` service in the v0.7 release.  

Would you mind taking a pass at relocating the related structures under `authorize` and removing it from the proxy service settings?

We could consider this a minor version change.

travisgroth:
This LGTM; just needs a rebase

travisgroth:
>Do you think it is worth removing the service prefix since it is both in authorize and authentication services

Right now we do have most settings broken down by the service that "owns" it and, as you point out, that has been a moving target.  I think that's worth considering across the board for an upcoming refactor.  

Either way, I think we're looking at a breaking change here if you're relying on a generated key.  I don't believe external keys or user provided keys will be impacted.  The upgrade notes can certainly be fleshed out a bit more to help.

Issue #88, opened 2020-04-27
Insecure is still using port 443 and https naming convention 


Vad1mo:
#83 introduced insecure. 

Unfortunately it is incomplete. For example when using together with Traefik, then Traefik is [guessing](https://docs.traefik.io/routing/providers/kubernetes-ingress/#communication-between-traefik-and-pods) based on port number 443  and the port (name?) if the service should called plain or over tls. 

The setups with HTTP traffic over port 443 confuses Traefik as it assumes by port number & port name that the upstream service is secure which it isn't. 

If `insecure=true` change Ports from 443 to 80 and rename port from https to http.

travisgroth:
@Vad1mo I've opened up #103 to try to address this.  Can you give it a shot and confirm it works as expected?

PR #103, opened 2020-05-15
Update port names and numbers for insecure mode

travisgroth:
Closes https://github.com/pomerium/pomerium-helm/issues/88

When using insecure mode, change port names and numbers.  While mostly cosmetic, this can assist ingress controllers and service mesh systems detect correct protocols automatically.


Vad1mo:
the services are still http + port 443

See this example:

```yaml
kind: Service
apiVersion: v1
metadata:
  name: pomerium-proxy
  namespace: pomerium
  selfLink: /api/v1/namespaces/pomerium/services/pomerium-proxy
  uid: dd6a4567-cb3b-4161-b953-7614e406f3e7
  resourceVersion: '16537130911'
  creationTimestamp: '2020-05-15T06:19:47Z'
  labels:
    app.kubernetes.io/component: proxy
    app.kubernetes.io/instance: pomerium
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: pomerium-proxy
    helm.sh/chart: pomerium-9.1.0
  annotations:
    meta.helm.sh/release-name: pomerium
    meta.helm.sh/release-namespace: pomerium
spec:
  ports:
    - name: http
      protocol: TCP
      port: 443
      targetPort: http
```

travisgroth:
Ah.  I forgot to mention this.  

The external service port is still defined explicitly by `service.externalPort`.  Try setting `service.externalPort` to `80`.

I'll see about making that default dynamic.  For now we're just relying on the explicit option.

travisgroth:
Actually, I think I have it.  Try again with just `insecure`.

Vad1mo:
@travisgroth tested your new code. It works as expected.

travisgroth:
> Maybe a readme update is needed:
> 
> The default value in `service.externalPort` should be `80` or `443` depending on `config.insecure`= `false` or `true`

Yep agreed.  I wanted to confirm before updating docs.  Thanks for testing!

PR #106, opened 2020-05-18
Insecure mounting secrets fix

### LINKED ISSUE #105: (feat): disable secrets mounting with config.insecure=true flag
Hello,
`--set config.insecure=true ` flag should disable also https://github.com/pomerium/pomerium-helm/blob/master/charts/pomerium/templates/authenticate-deployment.yaml#L125-L143 tls secrets mounting, its not possible to start deployment with `--set config.insecure=true` flag.

Thanks

lukasmrtvy:
Insecure mounting secrets fix #105

### DIFF
--- a/charts/pomerium/templates/secret.yaml
+++ b/charts/pomerium/templates/secret.yaml
@@ -54,9 +54,9 @@ stringData:
 
 {{- end -}}
 {{- if and .Values.forwardAuth.enabled .Values.forwardAuth.internal }}
-    forward_auth_url: https://{{ template "pomerium.proxy.fullname" . }}.{{ .Release.Namespace }}
+    forward_auth_url: {{ .Values.config.scheme }}://{{ template "pomerium.proxy.fullname" . }}.{{ .Release.Namespace }}
 {{ else }}
-    forward_auth_url: https://{{ template "pomerium.forwardAuth.name" . }}
+    forward_auth_url: {{ .Values.config.scheme }}://{{ template "pomerium.forwardAuth.name" . }}
 {{- end -}}
 {{- if .Values.config.policy }}
     policy:
--- a/charts/pomerium/templates/secret.yaml
+++ b/charts/pomerium/templates/secret.yaml
@@ -53,6 +53,11 @@ stringData:
 {{- end -}}
 
 {{- end -}}
+{{- if .Values.config.insecure }}
+    {{- $_ := set .Values.config "scheme" "http" }}
+{{ else }}
+    {{- $_ := set .Values.config "scheme" "https" }}
+{{- end }}
 {{- if and .Values.forwardAuth.enabled .Values.forwardAuth.internal }}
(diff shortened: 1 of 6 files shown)


lukasmrtvy:
This is problematic too: https://github.com/pomerium/pomerium-helm/blob/master/charts/pomerium/templates/secret.yaml#L56-L60  ( hardcoded scheme, will not work `config.insecure=true`  )

charts/pomerium/templates/secret.yaml
```diff
@@ -53,10 +53,15 @@ stringData:
 {{- end -}}
 
 {{- end -}}
+{{- if .Values.config.insecure }}
```
travisgroth, line 56: Can we move this into a named template in `_helpers.tpl` and reference that with `include`?  I don't want to be setting inline in a template.

It should also be an `internalScheme` or something.  The scheme for things like the authenticate service and an external forward auth endpoint should be independent.  Also see other comment on this.

charts/pomerium/templates/secret.yaml
```diff
@@ -53,10 +53,15 @@ stringData:
...
+    {{- $_ := set .Values.config "scheme" "http" }}
+{{ else }}
+    {{- $_ := set .Values.config "scheme" "https" }}
+{{- end }}
 {{- if and .Values.forwardAuth.enabled .Values.forwardAuth.internal }}
-    forward_auth_url: https://{{ template "pomerium.proxy.fullname" . }}.{{ .Release.Namespace }}
+    forward_auth_url: {{ .Values.config.scheme }}://{{ template "pomerium.proxy.fullname" . }}.{{ .Release.Namespace }}
 {{ else }}
-    forward_auth_url: https://{{ template "pomerium.forwardAuth.name" . }}
+    forward_auth_url: {{ .Values.config.scheme }}://{{ template "pomerium.forwardAuth.name" . }}
```
travisgroth, line 64: I think this should be `https` unless it is internal and insecure.  If the URL is being hit externally, it should be secure.

travisgroth:
> This is problematic too: https://github.com/pomerium/pomerium-helm/blob/master/charts/pomerium/templates/secret.yaml#L56-L60 ( hardcoded scheme, will not work with `config.insecure=true` )

Typically, forward auth endpoint is exposed via ingress so it would be `https` regardless.  I don't think it should be based purely on `config.insecure`.

> Btw, its little bit schizophrenic, something is set in config.yaml, something via env variables.

Yes, we've been slowly gravitating towards a consolidated secret with env as overrides.  It has been gradual but I expect to make the shift fully in the near future.

> https://github.com/pomerium/pomerium-helm/blob/master/charts/pomerium/templates/authenticate-deployment.yaml#L69-L72
> ,some urls have scheme prefix, some urls dont.

Yes, the defaults have the URL prefixed.

travisgroth:
Closed by #116

PR #108, opened 2020-05-25
Rootless

alemairebe:
This PR force the container to run as nonroot without any specific privileges.
I think if something can run without privileges, it should.


Vad1mo:
why the env with the ports 8443 or is because of the low ports? 
This solution won't work with insecure mode. 
Could you rebase on the current 9.1.0?

travisgroth:
Closing due to inactivity

PR #112, opened 2020-05-31
set correct version for helm chart

Vad1mo:
the version set was not 0.9.0 but still 0.8.0


travisgroth:
Hey @Vad1mo thanks for this.  We also need to bump the operator version to ensure configuration handling is functional.  Would you mind updating this to include it?  https://github.com/pomerium/pomerium-operator/releases/tag/v0.0.3 Thanks!

travisgroth:
resolved in #113

PR #115, opened 2020-06-03
Running policy through helm's template function

bjoernw:
Running the policy through helm's tpl function would allow for install-time overrides for policy parameters.

For example, if all your "from" urls have a base url that changes in every environment it would be useful to pass something like `cluster.baseUrl` as a --set parameter.


travisgroth:
Hi @bjoernw - LGTM but would you mind rebasing?  Thanks!

PR #116, opened 2020-06-04
Reorganize shared/common deployment settings

travisgroth:
This PR is an attempt to consolidate shared settings into the config yaml file out of ENV vars.  This allows logic to be centralized and permits further breakdown of templating.

Details:

- Move all current ENV vars to static config section, except the `SERVICES` var
- Move volumeMounts to a new named template
- Move common volumes to a new named template
- Move service-specific volumes to a new named template
- Derive scheme for `authorize` and `cache` URLs from templates
- Ensure we're setting the `grpc_address` correctly
- Derive probe scheme from templates
- Mark TLS related volumes optional to support insecure mode more easily

NB: the way `$ctx` is constructed to pass down a service name to `pomerium.volumes.service` feels gross.  Not sure if there's a cleaner way.

/cc @Vad1mo @lukasmrtvy as this fixes some (all?) issues with insecure logic.

### DIFF
--- a/charts/pomerium/README.md
+++ b/charts/pomerium/README.md
@@ -16,6 +16,7 @@
     - [Self Provisioned](#self-provisioned-1)
   - [Configuration](#configuration)
   - [Changelog](#changelog)
+    - [10.0.0](#1000)
     - [8.5.5](#855)
     - [8.5.1](#851)
     - [8.5.0](#850)
@@ -28,6 +29,7 @@
     - [3.0.0](#300)
     - [2.0.0](#200)
   - [Upgrading](#upgrading)
+    - [10.0.0](#1000-1)
     - [8.0.0](#800-1)
     - [7.0.0](#700-1)
     - [5.0.0](#500-1)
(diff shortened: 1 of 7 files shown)


charts/pomerium/templates/_helpers.tpl
```diff
@@ -299,8 +299,23 @@ https
 {{- end -}}
 {{- end -}}
 
+{{/*Creates static configuration yaml */}}
 {{- define "pomerium.config.static" -}}
 address: ":{{ template "pomerium.trafficPort.number" }}"
+grpc_address: ":{{ template "pomerium.trafficPort.number" }}"
+certificate_file: "/pomerium/cert.pem"
+certificate_key_file: "/pomerium/privkey.pem"
+certificate_authority_file: "/pomerium/ca.pem"
```
lukasmrtvy, line 308: Is this really needed in insecure mode? 
It means that You have to use `--set config.generateTLS=true` flag. 
Related: https://github.com/pomerium/pomerium-helm/pull/106/commits/0b83c46b62c852395c3eb13c1a67f3193085ab7b

charts/pomerium/README.md
```diff
@@ -298,6 +304,32 @@ A full listing of Pomerium's configuration variables can be found on the [config
...
+  Some of the impacted chart values are listed below:
+
+  - `authenticate.idp.provider`
+  - `authenticate.idp.url`
+  - `authenticate.cacheServiceUrl`
+  - `authenticate.idp.scopes`
+  - `config.insecure`
+  - `proxy.authenticateServiceUrl`
+  - `proxy.authorizeInternalUrl`
+  - `config.rootDomain` - Impacts ENV vars:
```
desimone, line 321: Typo?  ` - Impacts ENV vars:` looks dangling?

travisgroth: No...but maybe should rethink how to convey that.

PR #120, opened 2020-07-03
Prevent Helm output error when using environment variables to set identity provider credentials

### LINKED ISSUE #119: Error shown in Helm output when environment variables are used for identity provider credentials
## Overview

I use environment variables passed via `extraEnv` to set the identity provider (GitLab) credentials: `IDP_CLIENT_ID` and `IDP_CLIENT_SECRET`. The reason for using environment variables rather than `authenticate.idp.{clientID, clientSecret}` is that my credential values are stored in Vault, and are accessed by the Pods dynamically using the `bank-vaults` webhook mechanism:

```
extraEnv:
  IDP_CLIENT_ID: vault:ci/data/sandbox/pomerium#IDP_CLIENT_ID
  IDP_CLIENT_SECRET: vault:ci/data/sandbox/pomerium#IDP_CLIENT_SECRET
```

The functionality works fine, but the Helm chart outputs an error message after deploying the chart, implying that something is wrong, when there isn't actually anything wrong.

Mentioning my use of `bank-vaults`-style value references is only relevant for the purpose of explaining my need for using environment variables; this also applies to hard-coded values set in the `extraEnv` block.

## Actual Behavior

1. I deploy the Helm chart with identity provider credentials defined in environment variables.
1. The Pods spin up successfully.
1. Pomerium successfully provides Forward Auth functionality for my ingress.
1. The `helm install` command exits cleanly, but outputs the following error message:
   ```
   NOTES:
   ##############################################################################
   ####        ERROR: You did not set a valid identity provider              ####
   ##############################################################################

   This deployment will be incomplete until you configure a valid identity provider.

   Currently supported providers:

       - Okta
       - Google
       - Azure Active Directory
       - OneLogin

   See the values.yaml file to see what values are required for each provider.

   If you are having trouble with the configuration of a provider please visit
   the official documentation:

       https://www.pomerium.io/docs/identity-providers.html
   ```

The only way to avoid this error is to define "dummy" `authenticate.idp.{clientID, clientSecret}` values, and rely on the hierarchical priority given to the real credential values set by environment variables:

```
authenticate:
  idp:
    provider: gitlab
    clientID: DEFINED_AS_ENV_VAR
    clientSecret: DEFINED_AS_ENV_VAR

extraEnv:
  IDP_CLIENT_ID: <vault specification key, or hardcoded credential value>
  IDP_CLIENT_SECRET: <vault specification key, or hardcoded credential value>
```

## Expected Behavior

The Helm chart should allow me to use environment variables instead of `values.yaml` configuration keys for specifying identity provider credentials, without implying that there is an error in my configuration.

## Steps to Reproduce the Problem

1. Use the Helm chart to deploy a working implementation of Pomerium, defining the identity provider credentials via `authenticate.idp.{clientID, clientSecret}` in the `values.yaml` file:
   ```
   authenticate:
     idp:
       provider: gitlab
       clientID: aaaabbbbccccdddd1111222233334444
       clientSecret: ddddccccbbbbaaaa4444333322221111
   ```
1. Notice that no error message is generated.
1.  Use the Helm chart to deploy the same implementation of Pomerium, but define the identity provider credentials via environment variables instead:
   ```
   authenticate:
     idp:
       provider: gitlab

   extraEnv:
     IDP_CLIENT_ID: aaaabbbbccccdddd1111222233334444
     IDP_CLIENT_SECRET: ddddccccbbbbaaaa4444333322221111
   ```
1. Notice that an error message is displayed in the `helm install` output.

## Specifications

- Chart version: `v10.0.1`
- Pomerium version: `v0.9.2`
- K8s version: `v1.15.11`

## Rectification

Modify `"pomerium.providerOK"` in `_helpers.tpl` to also check for identity provider credentials having been set via environment variables.

davidrosson:
## What

- Update `pomerium.providerOK` in `_helpers.tpl` to consider identity provider credentials set via environment variables as valid (to prevent generating an invalid error message).
- Update the list of valid identity providers that is shown, in the case that invalid identity provider credentials actually are provided.

## Why

- Fixes #119
- An error message should not be displayed when valid input is provided.

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 10.0.2
+version: 10.0.3
 appVersion: 0.9.2
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo-long.svg
--- a/charts/pomerium/templates/NOTES.txt
+++ b/charts/pomerium/templates/NOTES.txt
@@ -36,9 +36,12 @@ This deployment will be incomplete until you configure a valid identity provider
 
 Currently supported providers:
 
-    - Okta
-    - Google
     - Azure Active Directory
+    - AWS Cognito
+    - GitHub
+    - GitLab
+    - Google
+    - Okta
     - OneLogin
 
 See the values.yaml file to see what values are required for each provider.
--- a/charts/pomerium/templates/_helpers.tpl
+++ b/charts/pomerium/templates/_helpers.tpl
@@ -140,6 +140,8 @@ Adapted from : https://github.com/helm/charts/blob/master/stable/drone/templates
 {{- if .Values.authenticate.idp -}}
   {{- if .Values.config.existingSecret -}}
   true
+  {{- else if and .Values.extraEnv.IDP_CLIENT_ID .Values.extraEnv.IDP_CLIENT_SECRET -}}
+  true
   {{- else if eq .Values.authenticate.idp.clientID "" -}}
   false
   {{- else if eq .Values.authenticate.idp.clientSecret "" -}}


davidrosson:
If #122 is accepted, then this "general" fix for #119 will not be needed, as #122 provides a cleaner solution (injecting the credential-containing environment variables into _only_ the `authenticate` Pods), to address the needs described in #121.

davidrosson:
Superseded by #122

PR #122, opened 2020-07-03
Enable 'authenticate'-only environment variables and ServiceAccount

### LINKED ISSUE #121: Allow injecting specific environment variables into only the 'authenticate' Pods
## Overview

I use environment variables passed via the existing `extraEnv` capability to set the identity provider (GitLab) credentials: `IDP_CLIENT_ID` and `IDP_CLIENT_SECRET`. The reason for using environment variables rather than `authenticate.idp.{clientID, clientSecret}` is that my credential values are stored in Vault, and are accessed by the Pods dynamically using the `bank-vaults` webhook mechanism:

```
extraEnv:
  IDP_CLIENT_ID: vault:ci/data/sandbox/pomerium#IDP_CLIENT_ID
  IDP_CLIENT_SECRET: vault:ci/data/sandbox/pomerium#IDP_CLIENT_SECRET
```

This functionality currently works fine. However, the only way to set environment variables in Pods is to use this "global" `extraEnv` capability, which injects the defined environment variables into _all_ components' Pods (`authenticate`, `authorize`, `cache`, `proxy`). Since the credentials are only needed by the `authenticate` Pods, it would be useful to be able to set these environment variables _only_ for those Pods that require them, e.g.:

```
authenticate:
  deployment:
    extraEnv:
      IDP_CLIENT_ID: vault:ci/data/sandbox/pomerium#IDP_CLIENT_ID
      IDP_CLIENT_SECRET: vault:ci/data/sandbox/pomerium#IDP_CLIENT_SECRET
``` 

Extending my use of environment variables for credential injection, it would be nice if the Helm deployment could generate an optional Kubernetes `ServiceAccount` to be used by the `authenticate` deployment (only if RBAC is enabled). This would allow scoping access to Vault secrets via a policy/auth configuration bound to the generated `ServiceAccount`.

davidrosson:
## What

- Enables setting environment variables specific to the `authenticate` deployment.
- Creates a Kubernetes ServiceAccount for the `authenticate` deployment when RBAC is enabled.
- Recognizes identity provider credentials set via environment variables, and doesn't display an error about "invalid" identity provider credentials.
- Updates the list of valid identity providers.

## Why

- Fixes #121.
- Enables usage of Vault to cleanly inject protected identity provider credentials via environment variables.
- Prevents an incorrect error message from being displayed when valid input is provided (more details in #119).
- Provides an `authenticate`-specific fix, improving upon the "general" fix proposed by #120.


davidrosson:
@travisgroth Thanks for the feedback. Here's where I'd be interested in some clarification regarding next steps:

1. Regarding the isolation of IdP credentials to just the service(s) requiring them, I had assumed that only the `authenticate` service required them since this is where they are currently specified in the default `values.yaml` file (`authenticate.idp.{clientID,clientSecret}`). Also, `authenticate.go` in the Pomerium source code is the only place I could find where the config options are validated for the existence of IdP credentials. In any case, if it simplifies the chart, I'm not opposed to just specifying them via the existing `extraEnv` mechanism, instead of introducing a new `authenticate`-specific `extraEnv` option. Should I go ahead and drop the new `extraEnv` suggestion from the PR?

1. I agree with the idea that a ServiceAccount could also be useful for other things in the future (beyond just my Vault use case), and therefore it would be worth applying it to all the components. Following the convention in the chart's templates dir, all of the component-specific template files are getting prefixed with the component name (e.g. `authenticate-deployment.yaml`), and any of the generic template files are just named after the resource type (e.g. `role.yaml`). I would like to create a component-agnostic template file called `serviceaccount.yaml` for the ServiceAccount creation, but there is already a template file with that name, which is used by the `pomerium-operator`. Would it make sense to prefix all of the `operator`-specific template files with `operator-`, to make their function clearer and free up the `serviceaccount.yaml` name? `operator-deployment.yaml` already does this, but e.g. `configmap.yaml` (only used by `operator`) does not.

travisgroth:
Sorry for the long pause here - some other things have come up that has caused me to rethink my original feedback and I wanted to make sure the changes here make sense with an eye to the future.

>Regarding the isolation of IdP credentials to just the service(s) requiring them...

There are upcoming changes that will share at least idp credentials between two services.  I think something that is becoming clear to me is that the chart `values` should not be service specific.  It makes changes to the underlying Pomerium security posture or even operations challenging.  I don't want to tackle that in this PR but just philosophically I think the user shouldn't care exactly which service uses which settings.  It eliminates confusion.

>I agree with the idea that a ServiceAccount could also be useful for other things in the future (beyond just my Vault use case), and therefore it would be worth applying it to all the components. Following the convention in the chart's templates dir, all of the component-specific template files are getting prefixed with the component name (e.g. authenticate-deployment.yaml), and any of the generic template files are just named after the resource type (e.g. role.yaml). 

I think the way those arise has to do with how nasty `if` blocks would be to manage the differences between components.  For instance, I think role/rolebindings could easily be one template file for all components since there's not much logic to it.  Deployments are really bad to share or become very long `if` blocks.

>I would like to create a component-agnostic template file called serviceaccount.yaml for the ServiceAccount creation, but there is already a template file with that name, which is used by the pomerium-operator. Would it make sense to prefix all of the operator-specific template files with operator-, to make their function clearer and free up the serviceaccount.yaml name? operator-deployment.yaml already does this, but e.g. configmap.yaml (only used by operator) does not.

With the above in mind, I think having a single `serviceaccount.yaml` is fine for all the components.  It's really just a list of service accounts to create.  

Finally, as to how I see this moving forward:

(1) I believe a service account per component actually does make sense.  In many deployments, you may wind up granting similar permissions to more than one, but there is some difference in privilege required for each service.  In cloud environments, operators may chose to bind multiple kubernetes service accounts to the same underlying iam service account, but those with more fine grained access control requirements can exercise them.  

(2) Adding in an extraEnv parameter for each service sounds reasonable.  I like it as an escape valve without much compatibility guarantees that I'd like to grant the normal values.  Please also include the same structure for extraEnvRef, as I'm sure that will come up at some point.

(3) I believe there should be a stub to put annotations on each service's individual account.  This is usually how cloud providers do bindings from kubernetes to iam.  We should enable precise binding on a per service basis in addition to the broad annotation value we currently pass in.

Hopefully that all makes sense.

davidrosson:
Thanks again for the feedback. I'm just returning from a few weeks away, and have implemented most of what we discussed.

I see that in the meantime you've created a ServiceAccount specification for the `authorize` component to facilitate that API proxy. Since my updates are applying a ServiceAccount to each component, you'll want to check that my changes haven't broken your intended use of the `authorize` ServiceAccount.

Also, I noticed that you implemented a "global" version of `extraEnvFrom`:

```
{{- if .Values.extraEnvFrom }}
        envFrom:
{{        toYaml .Values.extraEnvFrom | indent 8 }}
{{- end }}
```

My planned approach was to do something that would allow both "global" and component-specific `extraEnvFrom` usage, but I haven't implemented it in this PR because it would conflict with the way the new "global" approach has been implemented (meaning, the new approach only sets `envFrom:` if a "global" `extraEnvFrom` if present, and uses YAML injection), and my changes would potentially break functionality for users who are already using this.

I had planned to implement this as shown in the snippets below.

In each `<component>-deployment.yaml` file (this example is for `authenticate`):

```
        envFrom:
{{- range $name := .Values.extraEnvFrom.ConfigMapName }}
        - configMapRef:
            name: {{ $name }}
{{- end }}
{{- range $name := .Values.authenticate.deployment.extraEnvFrom.ConfigMapName }}
        - configMapRef:
            name: {{ $name }}
{{- end }}
{{- range $name := .Values.extraEnvFrom.SecretName }}
        - secretRef:
            name: {{ $name }}
{{- end }}
{{- range $name := .Values.authenticate.deployment.extraEnvFrom.SecretName }}
        - secretRef:
            name: {{ $name }}
{{- end }}
```

And in `values.yaml`:

```
authenticate:
  deployment:
    extraEnvFrom:
      ConfigMapName:
        - my-authn-configmap-1
        - my-authn-configmap-2
      SecretName:
        - my-authn-secret-1
        - my-authn-secret-2

extraEnvFrom:
  ConfigMapName:
    - my-global-configmap-1
    - my-global-configmap-2
  SecretName:
    - my-global-secret-1
    - my-global-secret-2
```

I also wanted to mention that I'm actually moving to a new company soon, so my last day with my current company is this coming Friday. As a result, after the next couple days I won't have access to the resources where I've been running and testing this, and likely won't have the spare time to work on this anymore. A couple guys from my current company might pick this up where I left off, if we're not able to resolve things in the next day or two.

travisgroth:
>I see that in the meantime you've created a ServiceAccount specification for the authorize component to facilitate that API proxy. Since my updates are applying a ServiceAccount to each component, you'll want to check that my changes haven't broken your intended use of the authorize ServiceAccount.

Yes, sorry I had to race ahead there.  We had a new feature that relied on having a non-default service account to work well.  I tried to keep it minimal so as not to step on any work you had in flight.  I'll take a look.

> My planned approach was to do something that would allow both "global" and component-specific extraEnvFrom usage...

Understood.  The intent with the way that was implemented was to not get too deep into the various config sources.  The `extra*` options being an escape valve, it's easier to give more control to the operator without too much checking.

> I also wanted to mention that I'm actually moving to a new company soon, so my last day with my current company is this coming Friday. As a result, after the next couple days I won't have access to the resources where I've been running and testing this, and likely won't have the spare time to work on this anymore. A couple guys from my current company might pick this up where I left off, if we're not able to resolve things in the next day or two.

I appreciate the work put in here.  I wish the timing was a little better, as we've been prepping for a release and I only now was able to get back to this.  It looks fairly close to ready so I think we can finish this up one way or another.  The functionality will be beneficial.  👍

travisgroth:
@davidrosson make sure you have `Allow edits and access to secrets by maintainers` checked on this PR so I can update the branch and maintain your contributions.

davidrosson:
Happy to contribute. Pomerium has been an excellent addition to our environments, and this Helm chart aligned really well with our existing management approach. Hopefully there will be a chance to use it again on my next project in the coming months.

travisgroth:
@davidrosson I rebased and took care of some minor tweaks/fixes around the service accounts.  The changes LGTM now.  Would you be able to document the new options in the README?

PR #142, opened 2020-08-28
Raise (commented out) CPU and memory limits.

rspier:
Too low limits can cause starvation and startup issues.  Suggest a higher
limit to prevent it.  See pomerium/pomerium#1337 for more details.

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 12.2.0
+version: 12.2.1
 appVersion: 0.10.0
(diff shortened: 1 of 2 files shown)


rspier:
Chart Test is failing because:

```
✖︎ pomerium => (version: "12.1.0", path: "charts/pomerium") > Chart version not ok. Needs a version bump!
```
Since this PR just modifies a comment, I don't think bumping the version is actually useful.

travisgroth:
@rspier while it is just a comment, I'm not 100% sure what the release automation will do without a version bump here.  Can you bump to `12.2.1` please?

rspier:
Done.

travisgroth (approved):

Issue #141, opened 2020-08-20
error 500 when set service.externalPort 80 and config.insecure true 


jbhuguenin:
authenticate always try to connect on cache svc with 443 port while svc is bind on 80

travisgroth:
Hi @jbhuguenin can you share your values (with any redeactions), chart version and image version?

Logs from authenticate would also be helpful.

hugocortes:
@jbhuguenin can you check if this pr fixes your issue? https://github.com/pomerium/pomerium-helm/pull/147

I ran into similar issue in my kubernetes cluster which is using istio with mtls. I resolved this issue by manually changing Authenticate and Proxy deployment port names to `http2` along with the corresponding service names. (or install using pr helm chart)

desimone:
Closing this due to inactivity. Please reopen if still an issue.

PR #147, opened 2020-09-09
fix: probes should only use http or https schemes

hugocortes:
Signed-off-by: Hugo Cortes <contact@hugocortes.dev>

## Summary

Proxy and Authenticate deployments will be rejected when providing `service.httpTrafficPort.nameOverride=http2` as per instructions when using istio with mtls. The error returned is:
```sh
spec.template.spec.containers[0].livenessProbe.httpGet.scheme: Unsupported value: "HTTP2": supported values: "HTTP", "HTTPS
```

This fix simply adds a `pomerium.httpTrafficPort.scheme` template which is dependent on `config.insecure` and will be used by readiness and liveness probe schemes

## Related issues

Possible fix for #141 
* I first experienced the same 500 error when setting `config.insecure=true` without the http2 name override. Awaiting confirmation that this chart fix resolves issue.

**Checklist**:
- [x] add related issues
~- [ ] update Configuration in README~ n/a
~- [ ] update Changelog in README (major versions)~ n/a
~- [ ] update Upgrading in README (breaking changes)~ n/a
- [x] ready for review

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 12.2.1
+version: 12.2.2
 appVersion: 0.10.0
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo-long.svg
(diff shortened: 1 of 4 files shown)


travisgroth:
Hi @hugocortes - Interesting.  I think that overriding that is no longer needed but no reason not to do the right thing here.  Thanks.

travisgroth (approved):

PR #166, opened 2020-10-19
Allow string templates in `config.policy` field

### LINKED ISSUE #153: config.policy should support multi-line templates
## What happened?

I defined the following in my values.yaml file:
```yaml
config:
  policy: |
      {{- range .Values.services }}
        - from: https://{{ .from }}
          to: {{ .to }}
      {{ toYaml .policy | indent 4 }}
      {{- end }}
```

`services` contains some home made structure to declare the various elements of the policy.

`pomerium.config.dynamic` is calling `{{ tpl (toYaml .Values.config.policy) . | indent 2 }}` and thus the resulting policy is:
```yaml
policy:
  |
  - from: https://xxxxx
    to: yyyyy
    etc…
```

As you can see, the `|` stayed because of the toYaml I believe.

## What did you expect to happen?

That the content of the `config.policy` is used directly as a template and as yaml.

I fixed the problem just by removing the `toYaml` operator in the `_helpers.tpl` file but I'm not sure it is not going to introduce regressions for other requirements :)

## What did you see in the logs?

Pomerium failed to parse the policy because it is expecting an array.

travisgroth:
## Summary
Parse the `config.policy` field so that either simple templates or full string templates can be used.

//CC @victornoel 

## Related issues
Closes #153 

**Checklist**:
- [x] add related issues
- [x] ready for review


calebdoxsey (dismissed):

charts/pomerium/README.md
```diff
@@ -216,7 +216,7 @@ A full listing of Pomerium's configuration variables can be found on the [config
...
-| `config.policy`                                              | List of routes and their policies.  [See more](https://www.pomerium.com/reference/#policy).                                                                                                                                                                                                        |                                                                             |
+| `config.policy`                                              | List of routes and their policies.  Can accept template values or string templates.  [See more](https://www.pomerium.com/reference/#policy).                                                                                                                                                       |                                                                             |
```
desimone, line 219:
```suggestion
| `config.policy`                                              | List of routes and their policies. Accepts template values or string templates.  [See more](https://www.pomerium.com/reference/#policy).                                                                                                                                                       |                                                                             |
```

desimone (dismissed):

calebdoxsey (approved):

Issue #167, opened 2020-10-21
pomerium-cache pod startup failing with redis: `invalid port ":%Adar5{&" after host`


stuartpb:
I've got `redis.enabled:true` in my Helm Release values, and my `pomerium-cache` pod is in CrashLoopBackOff:

```logs
[stuart@stushiba ~]$ kubectl logs pomerium-cache-56587db64-nln56 -n pomerium 
{"level":"info","version":"0.10.6-1601514118+a711475","time":"2020-10-21T05:18:59Z","message":"cmd/pomerium"}
{"level":"warn","time":"2020-10-21T05:18:59Z","message":"config: cache url will be deprecated in v0.11.0"}
{"level":"warn","Addr":":5443","GRPCAddr":":5443","time":"2020-10-21T05:18:59Z","message":"config: default http handler changed"}
2020/10/21 05:18:59 [INFO][cache:0xc000311f80] Started certificate maintenance routine
{"level":"info","port":"42531","time":"2020-10-21T05:18:59Z","message":"gRPC server started"}
{"level":"info","port":"36703","time":"2020-10-21T05:18:59Z","message":"HTTP server started"}
{"level":"warn","service":"directory","provider":"github","error":"unexpected end of JSON input","time":"2020-10-21T05:18:59Z","message":"invalid service account for github directory provider"}
{"level":"warn","provider":"github","time":"2020-10-21T05:18:59Z","message":"no directory provider implementation found, disabling support for groups"}
{"level":"warn","service":"databroker","error":"redis.DialURL(): parse \"rediss://:%Adar5{&?<>\\\"7ZDlSXMxSOFt\\\"!s4g?C\\\\@pomerium-redis-master.pomerium.svc.cluster.local:6379\": invalid port \":%Adar5{&\" after host","time":"2020-10-21T05:18:59Z","message":"failed to save server version."}
{"level":"info","time":"2020-10-21T05:18:59Z","message":"enabled cache service"}
{"level":"info","service":"identity_manager","time":"2020-10-21T05:18:59Z","message":"initializing directory groups"}
{"level":"info","addr":"127.0.0.1:42531","time":"2020-10-21T05:18:59Z","message":"starting control-plane gRPC server"}
{"level":"info","addr":"127.0.0.1:36703","time":"2020-10-21T05:18:59Z","message":"starting control-plane HTTP server"}
[...]

PR #168, opened 2020-10-21
Use randAlphaNum 10 for default Redis password

stuartpb:
This matches [the default value used by Bitnami's upstream Redis chart][1].

[1]: https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/_helpers.tpl#L231

Using non-alphanumeric characters in the Redis password is liable to introduce characters that break URL parsing in the databroker storage connection string.

## Related issues

Issues with URL-unsafe characters were observed in #167

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/charts/pomerium/templates/_helpers.tpl
+++ b/charts/pomerium/templates/_helpers.tpl
@@ -543,7 +543,7 @@ redis
 {{- else if .Values.config.sharedSecret -}}
 {{-  .Values.config.sharedSecret | sha256sum -}}
 {{- else -}}
-{{- randAscii 32 -}}
+{{- randAlphaNum 10 -}}
 {{- end -}}
 {{- end -}}
 


travisgroth:
Hi @stuartpb - thanks for the contribution.  Would you mind DCO signing and bumping the chart version?  I think this can be a patch.

stuartpb:
The thing is, as noted in https://github.com/pomerium/pomerium-helm/issues/167#issuecomment-713465769, this change on its own would only fix one part of an overall implementation that still ultimately won't work.

From what I can determine, the most comprehensive fix/workaround/hack to make this function would be to keep the redis password in a secret generated within the same template as the Pomerium secret with the configuration, and have that default to the looked-up existing value in that secret (if any) before trying to generate a new one.

stuartpb:
For that matter, though, if we're choosing to have the template auto-generate secret values in the case where the user hasn't provided them, we might as well generate a static value for `SHARED_SECRET` within the Helm chart, and then consolidate on using the sha256 sum of that shared secret to derive the default Redis password in all cases (instead of just the one where it's provided in the chart values, as is the case now).

travisgroth:
Ah, I missed some of the commentary on the original issue.  Let me give this some thought.

stuartpb:
Since the `SHARED_SECRET` and `COOKIE_SECRET` want 256 full bits of randomness, I'm honestly not sure they should be generated by the chart at all - the best (practical) RNG you can get in Sprig is `(randAscii 32 | b64enc)`, which my calculations indicate is more like 167 bits of randomness. I think I've seen other systems with non-seed-based secret generation handle this by having autogeneration be managed by a Job, but that would require a whole bunch of 

Really, though, the ultimate way to rectify this issue shouldn't be in the template at all: it should be for Pomerium to have some kind of key-derivation step in place (like even just a truncated sha256 hash) so that `SHARED_SECRET` can be an arbitrary sequence of bytes with any amount of entropy, and the responsibility for shrinking/stretching that entropy into 256 bits isn't left to the surrounding context (in other words, left to chance).

stuartpb:
Actually, after taking a look at the technique used for [TLS generation](https://github.com/pomerium/pomerium-helm/blob/master/charts/pomerium/templates/tls-secrets.yaml), I think I'll submit a new PR that will follows that approach, generating these values in a separate Secret that is only created when `Release.isInstall` is true and `config.generateSecrets` key is not `false`, and/or a `config.useExistingSecrets` (defaulting to `true`) that will attempt to read the existing values of secrets when generating values in the template (using the sha256 hash of the SHARED_SECRET as the basis for the Redis password, since that allows the password to be reproducible without needing to generate any new randomness during an upgrade to enable Redis)

PR #169, opened 2020-10-23
Refactor to derive secrets from existing secrets in cluster

stuartpb:
## Summary

As described in https://github.com/pomerium/pomerium-helm/pull/168#issuecomment-714837290, this uses the same conditional-generation conditions that the chart uses for TLS certificates to generate the shared secret and cookie secret, as well as reading the values back from the cluster on subsequent upgrades.

It does this, with significant refactoring to declare variables within the same file instead of in outside templates (as described in https://github.com/helm/helm/issues/6456), so that other secrets (such as the Redis password) may be derived from the auto-generated secret.

Note that I've tested this as far as making sure the templates render and validate, whether `redis.enabled` is true or false, but I haven't actually tested it out on my cluster.

## Related issues

#167, #168

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 13.1.0
+version: 14.0.0
 appVersion: 0.10.6
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo-long.svg
--- a/charts/pomerium/README.md
+++ b/charts/pomerium/README.md
@@ -18,6 +18,7 @@
   - [Redis Subchart](#redis-subchart)
   - [Configuration](#configuration)
   - [Changelog](#changelog)
+    - [14.0.0](#1400)
     - [13.0.0](#1300)
     - [11.0.0](#1100)
     - [10.2.0](#1020)
@@ -34,6 +35,7 @@
     - [3.0.0](#300)
     - [2.0.0](#200)
   - [Upgrading](#upgrading)
+    - [14.0.0](#1400-1)
     - [13.0.0](#1300-1)
     - [12.3.0](#1230)
     - [11.0.0](#1100-1)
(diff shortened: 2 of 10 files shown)


stuartpb:
Hmm, looks like the CI system's Helm isn't up to date? It's failing on the part with the `lookup` call.

travisgroth:
> Hmm, looks like the CI system's Helm isn't up to date? It's failing on the part with the `lookup` call.

Strange.  We should be on 3.3.1 https://github.com/pomerium/pomerium-helm/blob/ca748fec5e380eb9d8690e3f0f91ec9c54e36a68/.github/workflows/test.yaml#L18 unless chart-testing is integrated directly and way behind.

I'm a little hesitant to bring `lookup` in from a compatibility perspective to handle what started as a bit of a corner case.  We do presently test for helm2 as well, and I'm not sure how well systems like ArgoCD play with `lookup`.  Will review more throughly though; I haven't had the time yet.

stuartpb:
I was hoping there'd be a way to write the template where you can turn off `config.useExistingSecrets` and it would skip over the `lookup` call for compatibility with those earlier versions of Helm (and a version check could even be used to set the value's default) - I think this is possible in Go templating, but it'd mean wrapping the lookups in an `{{- if -}}`, which I'd avoided because I wasn't sure about how template variables are supposed to be scoped. Looking at some of the other examples out there as I was writing this, I'm pretty sure it should be doable, and cleaner

travisgroth:
@stuartpb my take here is this is a lot of logic (especially with compatibility) to avoid setting two simple parameters.  The TLS generation logic is painful, but worth it to ensure users can securely bootstrap without complicated external commands.  Providing random values for two parameters are a lot less of an ask.  I'm not sure it's a problem that needs to be solved, as it hasn't previously come up as an issue.

Will chime in on the original redis issue with some thoughts on resolving that in particular.

stuartpb:
>  Providing random values for two parameters are a lot less of an ask. I'm not sure it's a problem that needs to be solved, as it hasn't previously come up as an issue.

Without being able to rely on the stability of those two default-random fields across upgrades, systems like [GitOps Toolkit](https://toolkit.fluxcd.io/) will constantly roll/churn new Helm releases that do nothing but change the secrets in use.

travisgroth:
>Without being able to rely on the stability of those two default-random fields across upgrades, systems like GitOps Toolkit will constantly roll/churn new Helm releases that do nothing but change the secrets in use.

Right.  We ask the user to provide them currently; I think we continue to do so.  They're easy enough to provide.

PR #176, opened 2020-12-01
Setting a custom authenticate url by setting a subdomain

mohsen0:
## Summary

So in order to be able to deploy pomerium in different clusters, the pages need to be redirected to their own authenticate service deployed within the clusters.
for example, if all DNS services (including) authenticate service in the installation are under `*.<cluster name>.example.com` the authenticate service URL will be `authenticate.<cluster name>.example.com` by setting:

```yaml
ingress:
  authenticate:
    subdomain: "authenticate.prod"
```

## Related issues

**Checklist**:
- ~[] add related issues~
- ~[] update Changelog in README (major versions)~
- ~[] update Upgrading in README (breaking changes)~
- [x] ready for review


mohsen0:
helm2 broken since the Redis chart doesn't support helm2 anymore, it is nothing to do with this PR.

travisgroth:
HI @mohsen0,

If I understand the use case, I'd expect each install to just use a different `rootDomain` to support that behavior.  For your example there, you'd set `config.rootDomain` to `prod.example.com`.  Does that not work as expected?

That said, customizing the actual name of the authenticate service makes complete sense.  I think the parameter names would be slightly different in that context but the core changes still make sense.

charts/pomerium/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 13.1.3
+version: 13.1.4
```
travisgroth, line 3: Please bump minor here.  I think that best adheres to semver.

charts/pomerium/README.md
```diff
@@ -325,6 +325,7 @@ A full listing of Pomerium's configuration variables can be found on the [config
...
 | `ingress.tls.hosts`                                          | Override automatic ingress tls hosts list                                                                                                                                                                                                                                                          | `[]`                                                                        |
+| `ingress.authenticate.subdomain`                             | Setting a custom authenticate url by setting a subdomain                                                                                                                                                                                                                                           | `authenticate`                                                              |
```
travisgroth, line 328: I think the parameter for this should be `name` rather than `subdomain`.  So `ingress.authenticate.name`, etc.

charts/pomerium/values.yaml
```diff
@@ -208,6 +208,8 @@ service:
     nameOverride: ""
 
 ingress:
+  authenticate:
+    subdomain: "authenticate"
```
travisgroth, line 212: See previous comment.  `name` rather than `subdomain` makes sense.

charts/pomerium/templates/_helpers.tpl
```diff
@@ -401,7 +401,7 @@ certificate_file: "/pomerium/cert.pem"
 certificate_key_file: "/pomerium/privkey.pem"
 certificate_authority_file: "/pomerium/ca.pem"
 {{- end }}
-authenticate_service_url: {{ default (printf "https://authenticate.%s" .Values.config.rootDomain ) .Values.proxy.authenticateServiceUrl }}
+authenticate_service_url: {{ default (printf "https://%s.%s" (.Values.ingress.authenticate.name | default "authenticate") .Values.config.rootDomain ) .Values.proxy.authenticateServiceUrl }}
```
travisgroth, line 404: I think this value is inconsistent with the README and default values?

charts/pomerium/templates/ingress.yaml
```diff
@@ -52,7 +52,7 @@ spec:
               serviceName: {{ template "pomerium.authorize.fullname" . }}
               servicePort: {{ template "pomerium.httpTrafficPort.name" . }}
     {{- end }}
-    - host: {{ printf "authenticate.%s" .Values.config.rootDomain }}
+    - host: {{ printf "%s.%s" (.Values.ingress.authenticate.name | default "authenticate") .Values.config.rootDomain }}
```
travisgroth, line 55: I think this value is inconsistent with the README and default values?

charts/pomerium/values.yaml
travisgroth: Also, if there are defaults in the templates, I think this can be empty.

travisgroth (changes requested):
I think it may also make sense to consolidate the actual authenticate URL as a helper template.  It looks like we derive it into places now.  Thoughts?

travisgroth:
> helm2 checks are broken since the Redis chart doesn't support helm2 anymore, it is nothing to do with this PR.

👍  I think helm2 is now basically EOL so we can probably pull that out.  Will address separately.

mohsen0:
> HI @mohsen0,
> 
> If I understand the use case, I'd expect each install to just use a different `rootDomain` to support that behavior. For your example there, you'd set `config.rootDomain` to `prod.example.com`. Does that not work as expected?
> 
> That said, customizing the actual name of the authenticate service makes complete sense. I think the parameter names would be slightly different in that context but the core changes still make sense.

Sorry my example was not a good one, 

basically allowing something like `authenticate-uk.prod.example.com` and `authenticate-au.prod.example.com`

charts/pomerium/values.yaml
mohsen0: fixed

charts/pomerium/templates/_helpers.tpl
mohsen0: fixed

charts/pomerium/templates/ingress.yaml
mohsen0: fixed.

charts/pomerium/README.md
mohsen0: ok done

charts/pomerium/Chart.yaml
mohsen0: Done

mohsen0:
> I think it may also make sense to consolidate the actual authenticate URL as a helper template. It looks like we derive it into places now. Thoughts?

Sure let me do that

mohsen0:
@travisgroth can you have another look?

mohsen0:
> > helm2 checks are broken since the Redis chart doesn't support helm2 anymore, it is nothing to do with this PR.
> 
> 👍 I think helm2 is now basically EOL so we can probably pull that out. Will address separately.

I removed the helm2 checks

charts/pomerium/templates/_helpers.tpl
travisgroth, line 404: I think you can use the `pomerium.authenticate.hostname` template here too

mohsen0: done

travisgroth (dismissed):
LGTM

travisgroth:
Actions is having issues now.  Will merge once the tests pass.

mohsen0:
it can be merged now

travisgroth (approved):

PR #178, opened 2020-12-17
prevent checksum churn with extraEnv secrets

travisgroth:
## Summary
Prevent autogeneration (and thus checksum churn) of `shared_secret` and `cookie_secret` when they're being set by `extraEnv`.

## Related issues
n/a

**Checklist**:
- [ ] add related issues
- [ ] ready for review

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: pomerium
-version: 14.0.0
+version: 14.1.0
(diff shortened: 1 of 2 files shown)


adinhodovic:
Can we move this forward? Thanks!

stuartpb:
This was why I put #169 together so long ago, to avoid churn when Redis is enabled.

travisgroth:
This fell out of date but should be unnecessary with the changes in https://github.com/pomerium/pomerium-helm/pull/220.

PR #190, opened 2021-04-01
update to v2 since the redis subchart requires v2

nicgrayson:
## Summary
The redis subchart is set to v2 so you should bump this chart to v2


travisgroth:
@nicgrayson I think that is a holdover from helm2 support.  We can probably drop that at this point.

Would you mind:

* bumping the major number
* update change log reflecting drop of helm 2 support
* moving dependencies to `Chart.yaml`

Thanks!

nicgrayson:
@travisgroth done and done

travisgroth (changes requested):
Minor typo.  

Could you move the contents of `requirements.yaml` and `requirements.lock` into `Chart.yaml` and `Chart.lock` respectively?  That will put us into full compliance with api `v2`.

nicgrayson:
@travisgroth I think I got those updated correctly. Do you want to upgrade the redis chart? Looks like it's old.

travisgroth:
> @travisgroth I think I got those updated correctly. 

I think the `lock` data needs to be migrated too.

> Do you want to upgrade the redis chart? Looks like it's old.

Not against it - the upstream changelog needs a review to see if there's any compatibility problems.  I think they're two major versions ahead now.

nicgrayson:
Not sure why this closed. I've updated the PR with version 18 and the needed chart.lock file.

travisgroth:
FYI added a version bump to address https://github.com/bitnami/charts/issues/5033

travisgroth (approved):

PR #202, opened 2021-07-07
Setting a custom forward-auth url by setting a subdomain

mikelorant:


travisgroth:
Hi - I'm not clear on how this differs from the `pomerium.forwardAuth.name` template.  Can you provide more context?

mikelorant:
That also changes the resource name if using `pomerium.forwardAuth.name`, which means using a FQDN creates a problem with the object names. This is only about changing the ingress hostname, exactly how https://github.com/pomerium/pomerium-helm/pull/176 was done.

mikelorant:
Context of why this PR is needed, I am using the same root domain for many environments so needed `forwardauth-dev.example.org`, `forwardauth-test.example.org` and `forwardauth.example.org`. This was previously not configurable and hardcoded to `forward-auth`.rootDomain.

travisgroth:
> Context of why this PR is needed, I am using the same root domain for many environments so needed `forwardauth-dev.example.org`, `forwardauth-test.example.org` and `forwardauth.example.org`. This was previously not configurable and hardcoded to `forward-auth`.rootDomain.

That should already be possible with `forwardAuth.nameOverride=forwardauth-dev.example.org`.

> That also changes the resource name if using `pomerium.forwardAuth.name`, which means using a FQDN creates a problem with the object names. This is only about changing the ingress hostname, exactly how #176 was done.

Can you elaborate?  I just noticed [this](https://github.com/pomerium/pomerium-helm/blob/d4cf16f6029e71c8e3efbac99993c30c3fc63f13/charts/pomerium/templates/_helpers.tpl#L307), which appears to be a bug.  It would probably fail with the operator enabled and an overridden forward auth name.  Is that what you're running into?

mikelorant:
Have just done a diff compare between the implementation changes I did and the method you recommended.

You are correct that it will change all the host name as I needed, however there is one aspect I noticed that doesn't sit well with me:

```
# Source: pomerium/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: forwardauth-development.example.org
```

I am very much against having FQDN as Kubernetes object names and I am always concerned of naming restrictions preventing valid host names from working.

This is also referenced in the ClusterRoleBinding and RoleBinding. This leads to references to it as well in the Deployment.

Is this something we could clean up instead? I do think the prior solution that was used for the authentication ingress would be consistent which is why I copied it identically.

Knowing there is a solution using the method you described would meet my needs, however I do think there is still value in this PR.

Thoughts?

travisgroth:
> I am very much against having FQDN as Kubernetes object names and I am always concerned of naming restrictions preventing valid host names from working.

Yes this looks like just an error.  It should have been a separate override for the operator service account.
 
> Is this something we could clean up instead? I do think the prior solution that was used for the authentication ingress would be consistent which is why I copied it identically.

Yes I think the right thing to do is:

- Add a proper option for overriding the operator's service account name
- Reference that value instead when building the operator service account name and references

At that point, I believe the functionality will be as intended and should fit your use case.  
 
> Knowing there is a solution using the method you described would meet my needs, however I do think there is still value in this PR.

As-is, I believe this PR breaks other functionality and changes the current defaults.  It would be best to just fix the bug, as the existing template and value already work as intended.

Let me know if you're interested in contributing the fix.

travisgroth:
https://github.com/pomerium/pomerium-helm/issues/204

mikelorant:
I will make the changes you requested and create a new PR for issue #204.

We can close this one as a decline request.

mikelorant:
>     * Rename the forward auth value to something more rational like `forwardAuth.name`.  My guess is that value was originally meant for the service account override.

The concern I have with this approach is this is inconsistent with other parts of the existing implementation. When you look at #176 it added a specific ingress authentication value. Users will expect other host names to be altered the same way. Do we need to move the naming part out of ingress for both `authenticate` and `forwardAuth`? When looking at other Helm charts, the host names are nearly always defined in the ingress block.

The use of `nameOverride` and `fullnameOverride` are standard Helm practises and should only ever be used for changing Kubernetes object names. So while the method of using `nameOverride` worked, it was the type of option that did not make sense to be changed for the use case I had (which was only the host name).

We may have some confusion between us in terms of understanding how the change should be made, so I apologise if I am not being clear.

travisgroth:
> The concern I have with this approach is this is inconsistent with other parts of the existing implementation. When you look at #176 it added a specific ingress authentication value. Users will expect other host names to be altered the same way. Do we need to move the naming part out of ingress for both `authenticate` and `forwardAuth`? When looking at other Helm charts, the host names are nearly always defined in the ingress block.

I see what you're saying.  My current reasoning is that both of those names are important for Pomerium's own configuration, regardless of whether Ingress is in use.  In retrospect, those probably shouldn't be under ingress.  We've historically relied heavily on an ingress configuration but this is changing and we may need to rethink some configuration.

> The use of `nameOverride` and `fullnameOverride` are standard Helm practises and should only ever be used for changing Kubernetes object names. So while the method of using `nameOverride` worked, it was the type of option that did not make sense to be changed for the use case I had (which was only the host name).

I think this is a pattern that we've had for a while but hasn't been scrutinized.  We're building service URLs with parameters that set resource names in the kubernetes API.  It may not be a good practice.  Need to look more closely at what the changes would entail and if it's worth breaking.

> We may have some confusion between us in terms of understanding how the change should be made, so I apologise if I am not being clear.

You're making fair points.  I think this is a little more complicated to get right than it first appeared, unfortunately.

PR #209, opened 2021-07-28
Adding labels to Pomerium config secret

adamantal:
## Summary

While Pomerium's components (authenticate, authorize, proxy, databroker) can be easily discovered through label expressions, Pomerium's config secret is not.
The labels put on the secret is too generic, and using them:
```
app.kubernetes.io/instance=supertubes
app.kubernetes.io/managed-by=Helm
app.kubernetes.io/name=pomerium
helm.sh/chart=pomerium-21.1.1
```
would result in also these secrets:
```
antala@ANTALA-M-9CMU ~ % k get secrets -A -l app.kubernetes.io/instance=supertubes -l app.kubernetes.io/name=pomerium
NAMESPACE  NAME                                      TYPE     DATA   AGE
pomerium   pomerium-pomerium                         Opaque   1      22h
pomerium   pomerium-pomerium-databroker-client-tls   Opaque   3      25h
pomerium   pomerium-pomerium-signing-key             Opaque   1      22h
```

It would be nice to add custom labels on the secret generated by the Helm chart to find it more easily. Unfortunately the name of the secret can be varied based on various Helm values, so I'd prefer a label based solution.

## Related issues
No related issue.

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/charts/pomerium/README.md
+++ b/charts/pomerium/README.md
@@ -18,6 +18,7 @@
   - [Redis Subchart](#redis-subchart)
   - [Configuration](#configuration)
   - [Changelog](#changelog)
+    - [22.1.0](#2210)
     - [22.0.0](#2200)
     - [21.0.1](#2101)
     - [21.0.0](#2100)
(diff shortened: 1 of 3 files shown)


travisgroth:
Hi @adamantal - in general, I think this is fine but the name feels off.  I think `config.extraSecretLabels` might be a better fit.  Thoughts?  Is there any prior art for naming this type of option in other charts you've seen?

adamantal:
@travisgroth thanks for checking the PR. Yeah, `config.extraSecretLabels` seems a better choice. Let me update the patch.

travisgroth (approved):

Issue #211, opened 2021-07-29
authenticate.redirectUrl config takes no effect


adamantal:
## What happened?
The following config takes no effect: `authenticate.redirectUrl`

## What did you expect to happen?
To have effect.

## Steps to reproduce
Deploy the chart with that config configured to anything.

## What's your environment like?

- Chart version: 22.0.0 + #209
- Container image: v0.14.7
- Kubernetes version: 1.20
- Cloud provider: EKS
- Other details: -

## What are your chart values?

This is the config I'm referring to.
```values.yaml
...
authenticate:
  ...
  redirectUrl: ""
  ...
...
```

## What are the contents of your config secret?
Not relevant

## What did you see in the logs?
Not relevant

## Additional context
I checked Pomerium's code and the redirect URL stems from the following formula: `proxy.authenticateServiceUrl` + `/oauth2/callback` see the code [here|https://github.com/pomerium/pomerium/blob/master/authenticate/handlers.go#L114]. The config is a bit confusing (though it would take effect), and has no use. It worth considering to remove it from `values.yaml`.

travisgroth:
Thanks @adamantal.  This looks like it hasn't worked for a very long time.  I think you're right and we should just remove it at this point.

Issue #204, opened 2021-07-16
forwardAuth.nameOverride changes operator service account name


travisgroth:
## What happened?

The operator service account name is changed when an unrelated option is set.

## What did you expect to happen?

Not that.

## Steps to reproduce

set `forwardAuth.nameOverride` to any value.

## What's your environment like?

- Chart version: 21.1.1

## What are your chart values?

n/a

## What are the contents of your config secret?

n/a

## What did you see in the logs?

n/a

## Additional context

_Originally posted by @travisgroth in https://github.com/pomerium/pomerium-helm/issues/202#issuecomment-880594997_

PR #214, opened 2021-08-03
Fix inconsistent/broken override values

travisgroth:
## Summary

Addresses some interrelated inconsistencies and bugs in our values files:

1. Rename `ingress.authenticate.name`. That option is useful even when ingress is not enabled.
2. Rename `forwardAuth.nameOverride`. It is meant to rename a service URL and not a Kubernetes resource, so `nameOverride` isn't appropriate.
3. Fix the Operator pod service account to use dedicated annotations rather than those from the deployment.
4. Derive the Operator pod service account name from a dedicated `nameOverride` field.  Previously it was using `forwardAuth.nameOverride`, which is a bug.
5. Remove unused option `authenticate.redirectURL`.

## Related issues
Closes https://github.com/pomerium/pomerium-helm/pull/202
Fixes https://github.com/pomerium/pomerium-helm/issues/204
Fixes https://github.com/pomerium/pomerium-helm/issues/211

**Checklist**:
- [x] add related issues
- [x] ready for review


desimone (approved):
LGTM

PR #215, opened 2021-08-03
Enabled version checking and new structure for newer versions of the Ingress API

### LINKED ISSUE #213: Support networking.k8s.io/v1 Ingress API
**Is your feature request related to a problem? Please describe.**

For Kubernetes clusters >= 1.19 (e.g. now we are running 1.21) one would get a deprecation notice that the currently used Ingress API by Pomerium `networking.k8s.io/v1beta1` is deprecated and will be removed in k8s v1.22.

**Describe the solution you'd like**

I would like to see Pomerium support the newest stable version `networking.k8s.io/v1` for cluster versions >= 1.19. That also has an implication on the structure of the Ingress object (e.g. using `service.name` instead of `serviceName` or the need for the `pathType`, .. etc.) 

**Describe alternatives you've considered**

No real alternatives .. I mean starting k8s v1.22 the Pomerium Ingress as is won't work anymore.

**Explain any additional use-cases**

Nothing further.

**Additional context**

Nothing further. If that would be a fix on the level of the charts only, I am happy to submit a PR myself.

sherifkayad:
Enabled version checking and new structure for newer versions of the Ingress API

## Summary

Currently the Pomerium Ingress doesn't support Ingress `networking.k8s.io/v1`. This PR is to enable that with backwards compatibility for older versions of the Ingress API / k8s Clusters using Pomerium.

## Related issues

Fixes #213 

**Checklist**:
- [x] add related issues
- [x] ready for review


sherifkayad:
@travisgroth as discussed on #213 could you please review this one?

charts/pomerium/templates/_helpers.tpl
```diff
@@ -638,11 +638,20 @@ true
 Return the appropriate apiVersion for ingress.
 */}}
 {{- define "ingress.apiVersion" -}}
-{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" -}}
-{{- print "networking.k8s.io/v1beta1" -}}
-{{- else -}}
-{{- print "extensions/v1beta1" -}}
+  {{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= 1.19-0" .Capabilities.KubeVersion.Version) -}}
```
travisgroth, line 641: For my own information, why is the semver compare needed?  I'd expect just the v1 api being present is enough but I'm guessing this is here for a reason.

travisgroth (changes requested):
Looks like `pathType` needs to be specified now.

One question but generally LGTM.

sherifkayad:
@travisgroth sorry for the delay on this one. I will take it over today and re-attempt to fix the failing tests and your review comments. I believe the check on the version and capabilities is just there for making sure 100% that the new Ingress API version is there.

sherifkayad:
@travisgroth Can you have another look? I tested locally and seems to be fine this time (also applying the result to a cluster).

charts/pomerium/templates/_helpers.tpl
sherifkayad: Fixed

sherifkayad:
@travisgroth All checks have passed. I guess you need to sign off on the review in order to have that one merged 😉

charts/pomerium/templates/_helpers.tpl
```diff
@@ -638,11 +638,27 @@ true
...
+*/}}
+{{- define "ingress.isStable" -}}
+  {{- eq (include "ingress.apiVersion" .) "networking.k8s.io/v1" -}}
 {{- end -}}
+
+{{/*
+Return if ingress supports pathType.
+*/}}
+{{- define "ingress.supportsPathType" -}}
+  {{- or (eq (include "ingress.isStable" .) "true") (and (eq (include "ingress.apiVersion" .) "networking.k8s.io/v1beta1") (semverCompare ">= 1.18-0" .Capabilities.KubeVersion.Version)) -}}
```
travisgroth, line 661: Reviewed the API spec and for my own future self: this field was added in v1beta1 on 1.18, but v1beta1 was around prior.

charts/pomerium/values.yaml
```diff
@@ -214,8 +215,6 @@ service:
     nameOverride: ""
 
 ingress:
-  authenticate:
```
travisgroth, line 217: 😬 Thanks for catching this.

travisgroth (approved):

PR #216, opened 2021-08-05
Remove unnecessary quotation marks from static config

adamantal:
## Summary
The static part of the Pomerium config secret contains two config options the values of which is enclosed with quotation marks. Since they have leading `:` characters and not numbers, according to the YAML specification, the quotation marks are unnecessary.

Generally it shouldn't be an issue if we enclose a string value with quotation marks, but we have a Go-based automation that verifies the `checksum/secret` annotations (they indeed match the value of the secret). Consecutive `yaml.Unmarshal` and `yaml.Marhshal` calls on the secret remove that quotation mark (but leaves everything else as it is) - providing us with a different sha256sum. 

We'd kindly appreciate if you could accept this contribution, though it may only affect us and maybe the [pomerium-operator](https://github.com/pomerium/pomerium-operator).

## Related issues
None

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: pomerium
-version: 23.0.0
+version: 23.1.0
 appVersion: 0.14.7
 home: http://www.pomerium.io/
 icon: https://www.pomerium.com/img/logo-round.png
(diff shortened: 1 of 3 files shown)


travisgroth:
Thanks @adamantal that makes sense.

travisgroth (approved):

adamantal:
Thank you @travisgroth!

PR #219, opened 2021-09-16
Use cluster-internal name for internal forward auth

stuartpb:
I'm having trouble getting forward-auth working in my system, and the default internal forward-auth name seems wrong (like the kind of thing that would cause all forward-auth requests to come back with a 404). I don't know if this would fix the problems I've been having, but I'm not sure what context just the proxy name followed by the namespace works in.


travisgroth:
@stuartpb I'm curious - are you using nginx or traefik?  On the surface this is definitely a problem but, to my knowledge, the user's browser gets a redirect to the forward auth URL so I'm unclear on how the internal URL would actually work.

That said, this is more correct than it is currently implemented.  Would you mind bumping the version and signing your commit?

travisgroth:
Given [how long this has been implemented](https://github.com/pomerium/pomerium-helm/issues/89), I'm hesitant to make a breaking change.  This would be the first bug report about it that I can find, and it isn't conclusive if this fixes anything.  Closing pending additional feedback.

PR #220, opened 2021-10-06
Add Ingress Controller

travisgroth:
## Summary
- Replace Operator with Ingress Controller
- Refactor generated secrets to persist
- Improve use of external secrets for sensitive data such as IDP credentials; often from vault or another external operator.

## Related issues
Closes https://github.com/pomerium/pomerium-helm/issues/222

**Checklist**:
- [x] add related issues
- [x] ready for review


travisgroth:
Marking for review.  Will rebase/sign for DCO before merge.

wasaga:
if you set authenticate endpoint to proxied:

```yaml
authenticate:
  existingTLSSecret: "authenticate-ingress-sandbox-pomerium-io-prod"
  proxied: true
```

it does add a route to the `pomerium.config` (btw it's worth renaming `policy` to `routes`) 

```yaml
policy:
  - from: https://authenticate.ingress.sandbox.pomerium.io
    to: https://pomerium-authenticate.default.svc.cluster.local
    preserve_host_header: true
    allow_public_unauthenticated_access: true
    tls_server_name: authenticate.ingress.sandbox.pomerium.io
```

however, I would expect the cert from `existingTLSSecret` to be attached to the known certificates chain in proxy. It does not happen currently.

travisgroth:
> > however, I would expect the cert from existingTLSSecret to be attached to the known certificates chain in proxy. It does not happen currently.
> 
> Ah. I understand the confusion but it's for the authenticate service itself (eg you'd expect the SAN to be `pomerium-authenticate.[namespace].svc.cluster.local`); not the external URL.
> 
> Maybe clarifying in the README is the right thing to do? The certs in the proxy are not related to those on services behind it. You could use the same cert for both but I'm not sure we need to automatically optimize for that. Usually they're going to come from different CAs since one is internal and one is external.

@wasaga I added a new option - it behaves like extraTLSSecrets but is explicit and in a place people might expect the option to live.  https://github.com/pomerium/pomerium-helm/pull/220/commits/1b1b3eec9c25a51fac8387aa48e460d24c63d033

wasaga:
specifying `authenticate.existingExternalTLSSecret` makes it a chicken and egg problem : 

pomerium cannot start up if this certificate does not exist, 
and cert-manager cannot procure a certificate if http solver is used, as it requires an ingress controller up and running

travisgroth:
> specifying `authenticate.existingExternalTLSSecret` makes it a chicken and egg problem :
> 
> pomerium cannot start up if this secret does not exist,
> 
> ```
> {"level":"fatal","error":"config: options from config file \"/etc/pomerium/config.yaml\": validation error config: bad cert entry, base64 or file reference invalid. open /etc/pomerium/tls/authenticate-ingress-sandbox-pomerium-io-prod/tls.crt: no such file or directory","time":"2021-10-15T15:20:17Z","message":"cmd/pomerium"}
> ```
> 
> and cert-manager cannot procure a certificate if http solver is used, as it requires an ingress controller up and running
> 
> imho, requiring users to define an `Ingress` resource for `authenticate` is the best course of action.

We want proxied authenticate to work with and without the ingress controller enabled, but I agree there's an issue here.  If ingress controller is enabled maybe we can handle the situation differently.  Will take a closer look at that.

wasaga (approved):

PR #234, opened 2021-11-30
Don't unconditionally set certificate environment variables

### LINKED ISSUE #233: Chart does not function with insecure option enabled
When `config.insecure` is set to enabled, all services fail to start, giving "certificate file not found" errors.

From what I can tell, in the recent PR https://github.com/pomerium/pomerium-helm/pull/220, the template `pomerium.tls.internal.envVars` was added to `_helpers.tpl`. This helper includes the following environment variables:

```
- name: CERTIFICATE_FILE
  value: /pomerium/tls/tls.crt
- name: CERTIFICATE_KEY_FILE
  value: /pomerium/tls/tls.key
```

Instead of this helper being conditionally set when certificates exists, it is always included for the `authenticate`, `authorize` and `databroker` deployments, causing the services to fail to start, as they look for certificate files that does not exist.

DMarby:
Fixes #233

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: pomerium
-version: 25.0.3
+version: 25.0.4
 appVersion: 0.15.7
 home: http://www.pomerium.com/
 icon: https://www.pomerium.com/img/icon.svg
--- a/charts/pomerium/templates/authenticate-deployment.yaml
+++ b/charts/pomerium/templates/authenticate-deployment.yaml
@@ -59,7 +59,9 @@ spec:
 {{- end }}
 {{- end }}
         env:
+{{- if or ( and .Values.authenticate.tls.cert .Values.authenticate.tls.key ) .Values.authenticate.existingTLSSecret }}
 {{- include "pomerium.tls.internal.envVars" . | indent 8 }}
+{{- end }}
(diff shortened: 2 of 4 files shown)


travisgroth:
(closing/re-opening to kick the tests)

DMarby:
@travisgroth I've bumped the chart version now

travisgroth:
@DMarby tests are failing.  At a glance, I think the conditional isn't accounting for auto-generated certs

DMarby:
@travisgroth I took the conditional from the proxy deployment, are the certs handled differently per service? https://github.com/pomerium/pomerium-helm/blob/b62d8f081d8ae031d042675de4b1ef68de1cf8c6/charts/pomerium/templates/proxy-deployment.yaml#L64-L66

travisgroth:
> @travisgroth I took the conditional from the proxy deployment, are the certs being handled differently per service?
> 
> https://github.com/pomerium/pomerium-helm/blob/b62d8f081d8ae031d042675de4b1ef68de1cf8c6/charts/pomerium/templates/proxy-deployment.yaml#L64-L66

Sorry this took some time to get back to.

@DMarby the difference is with `proxy` we don't actually use the auto generated certs if they're created.  They're only for inter-service communication and the proxy won't pick them up.  

The rest of the services need to use them.  I think adding a check for `.Values.config.generateTLS` in addition to what you have here will do the trick.

Issue #235, opened 2021-12-17
Improved service mesh integration


travisgroth:
**Is your feature request related to a problem? Please describe.**

Presently, our `insecure` option changes a number of deployment time options, primarily optimized for running in a service mesh where all TLS is offloaded to the mesh.

However, as pointed out in https://github.com/pomerium/pomerium/issues/2387, our TCP tunneling does not work correctly if we don't terminate the HTTP connection ourselves.  More generally, users having to maintain service mesh configuration objects in _front_ of the pomerium proxy is additional configuration and performance overhead that is no longer necessary with the introduction of the ingress controller and our general pattern of being the user-facing edge proxy of any zero trust topology.

Concretely - our helm chart, as it is today, does not have all the right knobs to allow pomerium proxy to terminate inbound TLS/HTTP at edge, while offloading _internal_ communication to a secure service mesh.

**Describe the solution you'd like**

- Add configuration to enable/disable downstream TLS on the proxy service independently of the rest of the services
- Add per-deployment pod label customization settings
- Default proxy downstream TLS to enabled even in insecure mode
- Example istio deployment with new settings and labels as some istio specific labels are required

**Describe alternatives you've considered**

N/A

**Explain any additional use-cases**

Though the primary test case is istio, these settings should hopefully be useful in other service mesh deployments.

**Additional context**

n/a

travisgroth:
Related - https://github.com/pomerium/pomerium-helm/pull/234

anhdle14:
# Istio

For Istio ServiceMesh, one of the thing we can do is leverage the existing mTLS of Envoy pod via these tweaks to the pod itself.

```yaml
containers:
  - name: a-server
    ...
    volumeMounts:
      mountPath: /etc/certs/
      name: istio-certs
volumes:
  - emptyDir:
      medium: Memory
    name: istio-certs
```

And also with the annotations:

```yaml
spec:
  template:
    metadata:
      annotations:
        traffic.sidecar.istio.io/includeInboundPorts: ""   # do not intercept any inbound ports
        traffic.sidecar.istio.io/includeOutboundIPRanges: ""  # do not intercept any outbound traffic
        proxy.istio.io/config: |  # configure an env variable `OUTPUT_CERTS` to write certificates to the given folder
          proxyMetadata:
            OUTPUT_CERTS: /etc/istio-output-certs
[...]

PR #238, opened 2021-12-21
Add `insecureProxy` logic

### LINKED ISSUE #233: Chart does not function with insecure option enabled

travisgroth:
## Summary

Add the ability to run pomerium-proxy with TLS termination for downstreams while using insecure communication to other services.  The primary use case is when deployed at the edge of a service mesh handling TLS.  

When `config.insecure=true`, we try to derive the most logical configuration without breaking existing common patterns:

- `ingressController.enabled=false`: default to current behavior of insecure for all services.
- `ingressController.enabled=true` and `ingressController.config.operatorMode=true`: default to current behavior of insecure for all services.
- `ingressController.enabled=true`: run proxy in secure mode (terminating TLS) with the rest of the services insecure
- `config.insecure=true` and `config.insecureProxy=true`: run both proxy and internal services in insecure mode regardless of other settings

This PR also adds a handful of CI testing scenarios to make sure nothing is broken in several common configurations.  This would have prevented #233 .

## Related issues
implements most of #235 
Fixes #233 
Closes https://github.com/pomerium/pomerium-helm/pull/234

**Checklist**:
- [x] add related issues
- [x] ready for review


desimone (dismissed):
Code LGTM.

desimone (commented):
Code & text LGTM.

travisgroth:
@wasaga can you give this a review please.

wasaga:
I looked at templating code but there are too many interdependencies to be sure. I will have to test these parameter combinations.

wasaga (dismissed):
I can only confirm that this PR does not seem to break ingress controller: 

1. `ingressController.enabled=true`: checked, works as expected
2. `insecureProxy=true` is incompatible with ingress controller, probably it should complain if it's enabled. 
3. I do not understand `operatorMode=true` option and how is it compatible with ingress controller.

travisgroth:
> 3. I do not understand `operatorMode=true` option and how is it compatible with ingress controller.

This retains the old behavior of updating our running configuration without functioning as an actual ingress controller.  Eg, automated forward auth mode for another ingress controller.

> 2.  insecureProxy=true is incompatible with ingress controller, probably it should complain if it's enabled.

I've added a `validation.yaml` file that can check this.

wasaga (approved):

Issue #253, opened 2022-01-26
config.yaml is not idempotent when using existing shared secret


fculpo:
When using an existing shared secret with for instance COOKIE_SECRET and SHARED_SECRET set in it, the helm chart keeps remdomizing these values in the config.yaml.

Each apply with exactly the same values hence changes the config.yaml and is not idempotent.

I believe those lines are at fault
https://github.com/pomerium/pomerium-helm/blob/708f64ced7dbb1b398c6a8ced8a70035938830e7/charts/pomerium/templates/_helpers.tpl#L490

We should not set these when using an existing shared secret.

travisgroth:
Hi @fculpo,

You're right.  I think this was intended to be removed in https://github.com/pomerium/pomerium-helm/pull/220 but got lost in the changes.

PR #254, opened 2022-01-27
[pomerium] clean up `pomerium-cache` and shared secret location

travisgroth:
## Summary

In https://github.com/pomerium/pomerium-helm/pull/220, we intended to remove `shared_secret` and `cookie_secret` from the generated `pomerium` secret when we started storing them in the `pomerium-shared` secret.  While the latter takes precedence and works correctly, the original template still had the original secret generation that introduced churn in CD systems.  This PR removes the problematic lines.

While this is a bug, I've bumped the major version since this fix introduces a breaking change users may not have properly accounted for.

Since we're bumping the major for some cleanup, this PR also includes the final removal of the long deprecated `pomerium-cache` alias to the databroker.

## Related issues
Fixes https://github.com/pomerium/pomerium-helm/issues/253
https://github.com/pomerium/pomerium-helm/pull/220

**Checklist**:
- [x] add related issues
- [x] ready for review


travisgroth:
CLA seems stuck 🤷

wasaga (approved):

PR #285, opened 2022-03-25
pomerium: add support all-in-one pomerium service

clwluvw:
Signed-off-by: Seena Fallah <seenafallah@gmail.com>

## Summary
By enabling `pomerium.allInOne` we can have all services in one pod.

**Checklist**:
- [ ] add related issues
- [x] ready for review


clwluvw:
Hi @travisgroth 
Can you take a look at this PR?

travisgroth:
HI @clwluvw - could you expand on what the use case is for this change?  It adds a lot of conditions thus a lot of complexity.

travisgroth:
> > HI @clwluvw - could you expand on what the use case is for this change? It adds a lot of conditions thus a lot of complexity.
> 
> @travisgroth This helps to bring less complexity to the deployment and deploy all pomerium services all in one pod. yea it takes so much change but I believe this can bring less complexity to maintenance for some setups that don't want to deploy each service on a separate pod.

Can you elaborate on the maintenance burden?  Inter-service TLS comes to mind, though I think provisioning those certs is largely a solved problem,  so I'm curious what else that is directly exposed to the end user that leads to operational overhead.

clwluvw:
> Can you elaborate on the maintenance burden? Inter-service TLS comes to mind, though I think provisioning those certs is largely a solved problem, so I'm curious what else that is directly exposed to the end user that leads to operational overhead.

@travisgroth When each service comes into a single pod you should maintain more than 4 Deployments for your stability of your service (finding the bugs, gathering the logs and...) and so in small setups, this can leads to a big problem and all in one would be a much better deployment method for it (you only fight with one pod and maintain a single pod and...)!

clwluvw:
@travisgroth Any hope to support this type of deployment?

calebdoxsey:
@clwluvw Thanks for contributing. We do think there's value in an all-in-one mode simpler deployment for Kubernetes, especially for clusters with tighter resource requirements.

Our helm chart has become difficult to maintain because of so many options and deployment models. We're currently working through possible major changes to how the chart is written to make it easier to maintain, so we'd like to hold off on making the changes in this PR till we know how they might integrate with a future version of the helm chart.

Another possibility we're considering is to not use Helm at all for this simpler deployment model. Some products use a kustomize-friendly pre-built YAML file that can be applied directly to a Kubernetes cluster and then a user can customize by editing the ConfigMap of settings. (for example Argo CD uses this pattern: https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#non-high-availability)

Sorry for the delay, but it will probably be a while before we have a plan for moving forward.

wasaga:
https://github.com/pomerium/ingress-controller now has all-in-one deployment by default.

PR #289, opened 2022-03-31
ingress: Allow removal of the ingress TLS section

adinhodovic:
Allows disabling TLS if it's externally terminated (LB). Currently it's always defined, at least the hosts section.

https://github.com/pomerium/pomerium-helm/blob/main/charts/pomerium/templates/_helpers.tpl#L621-L641 (picks up ingress.hosts which I want defined).

LMK if it's any other way around it.

## Related issues

**Checklist**:
- [ ] add related issues
- [x] ready for review

### DIFF
--- a/charts/pomerium/Chart.yaml
+++ b/charts/pomerium/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: pomerium
-version: 31.0.2
+version: 31.0.3
 appVersion: 0.17.0
 home: http://www.pomerium.com/
 icon: https://www.pomerium.com/img/icon.svg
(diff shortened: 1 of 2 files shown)


travisgroth:
Hi @adinhodovic - can you explain the environment this is intended for?  Your ingress controller doesn't terminate TLS but the load balancer in front of it does?

adinhodovic:
> Hi @adinhodovic - can you explain the environment this is intended for? Your ingress controller doesn't terminate TLS but the load balancer in front of it does?

Yeah exactly, EKS with an ACM cert that the AWS LB uses - usually use letsencrypt + ingress but not for one of the setups I'm working at.

travisgroth:
> Yeah exactly, [EKS with an ACM cert that the AWS LB uses](https://kubernetes.github.io/ingress-nginx/deploy/#tls-termination-in-aws-load-balancer-nlb) - I usually use letsencrypt + ingress but not for one of the setups I'm working at.

Got it.  I think that's a simple enough change.  I think you're going to get a lint error - LGTM once tests are green.

travisgroth (approved):

PR #292, opened 2022-04-06
redis/fix: Fix default replica count

adinhodovic:
## Summary
I'd like only 1 replica, I'm aware I can pass the values themselves but if there's a default of 1 replica in the chart already, then I'd fix that instead.

Slave was changed to replica in https://github.com/bitnami/charts/pull/6102/files.

## Related issues

**Checklist**:
- [ ] add related issues
- [X] update Configuration in README
- [x] ready for review


travisgroth:
Thanks @adinhodovic.  Yes, changing the default makes sense.

travisgroth (approved):

PR #326, opened 2022-09-19
feat: Add hostnameDivider as an authenticate hostname option

adinhodovic:
## Summary

Just a simple change to adjust the hostname divider. Currently can't use dots but dashes, the DNS pattern is `authenticate-{rootdomain}`. Rootdomain in this case is `staging.example.com`, therefore I'd like the pomerium ingress to have the host `authenticate-staging.example.com` and not `authenticate.staging.example.com`.

The authenticate ingress always has the wrong domain: https://github.com/pomerium/pomerium-helm/blob/main/charts/pomerium/templates/authenticate-ingress.yaml#L31

**Checklist**:
- [x] add related issues
- [x] ready for review


desimone (dismissed):
LGTM 

\cc @wasaga

adinhodovic:
@wasaga can you take a look?

